PR #48, opened 2015-01-03
Enable safe by default

marshallswain:
This will also prevent the long list of notifications that show up upon connect.


mycaule:
Thanks for the patch. I tested it, everything seems ok.

PR #50, opened 2015-02-04
Reload the list of databases on every call

### LINKED ISSUE #34: collection list can not update when add a collection use mongo shell or other program
when i create a collection with mongo shell, the collection cannot find in the page /db/{database_name}.
in order to mangement the collection created by mongo shell, i need to restart mongo-express. 
it seems that server cache the collection names for every databse.

DavidePastore:
Fix #34.

### DIFF
--- a/app.js
+++ b/app.js
@@ -281,8 +281,12 @@ app.param('document', function(req, res, next, id) {
 //mongodb middleware
 var middleware = function(req, res, next) {
   req.adminDb = adminDb;
-  req.databases = databases; //List of database names
   req.collections = collections; //List of collection names in all databases
+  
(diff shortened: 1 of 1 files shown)


DavidePastore:
Hi! Is there anything wrong with this?

wulfsolter:
app.js in master has changed significantly since your PR, not sure why it wasn't merged at the time - I've just recently jumped in to clean up a bit. Looks like that section of code is now residing in router.js, https://github.com/andzdroid/mongo-express/blob/master/router.js#L143-L154

I think it's still great functionality, would appreciate a new PR though :)

PR #54, opened 2015-02-16
Make mongo-express works with cloudfoundry

ArthurHlt:
I've made some changes if you are interested. Your app is now available to works with a cloud platform (cloudfoundry) it does not affect your preexistent code it's just cloudfoundry compliant and can use services from him.


wulfsolter:
App has changed significantly since your PR. Would you like to resubmit?

PR #58, opened 2015-03-13
Complex queries

karthik25:
One thing I missed the most was to filter a collection based on more than one field and also the ability to limit the fields returned. This pull request takes care of that! 

Here is how the collection page looks like now (before entering a query):

without_query

Here is a screenshot with a query and the fields entered (along w/ the corresponding result):

with_query

If you have any suggestions do post them here. I am still looking in to whether I am missing anything else!

### DIFF
--- a/bson.js
+++ b/bson.js
@@ -54,6 +54,19 @@ exports.toBSON = function(string) {
   return sandbox.doc;
 };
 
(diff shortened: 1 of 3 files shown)


karthik25:
As a side note, the "Simple" tab has the form which was available before, I didn't modify any of that!

PR #65, opened 2015-05-23
mountable Express.js 4.x middleware

jeffreywescott:
This series of changes allows mongo-express to be mounted as a sub-app within another Express 4 app. It also improves the general modularity of the module such that the config object can be passed through as an argument to the middleware, the router, and all of the routes. This makes it so that the only file that depends on `config.js` is `app.js`, which won't be used when mounting as a middleware.


andzdroid:
Thanks for the pull request!

Issue #13, opened 2013-05-27
Hash-signs in document names are not properly url-encoded


oloflarsson:
I have a collection. In that collection there is a document called "rank#2"

Searching for it works fine:
http://localhost:8081/db/mstore/massiveperms_thing?key=_id&value=%23rank2
Note that the hash-sign is encoded to %23.

When I click it and try to open it does however not work:
http://localhost:8081/db/mstore/massiveperms_thing/#rank2

And if I manually enter
http://localhost:8081/db/mstore/massiveperms_thing/%23rank2
And edit and save I get the error "You forgot to enter a collection name!"

YvesMerl:
There is a general lack of URL encoding on all dynamic links.
Document ID with special URL characters will not be able to be opened, saved and deleted.

YvesMerl:
I made a pull request: https://github.com/andzdroid/mongo-express/pull/66

PR #66, opened 2015-05-29
Issue #13 : Fix document ID URL encoding on get, put and delete

YvesMerl:
I used the `url_encode` swig filter to URL encode document ID on **delete** and **put**.
And I used the native `encodeURIComponent` JavaScript function in `loadDocument(id)` to fix the **get**. 

Related to issue https://github.com/andzdroid/mongo-express/issues/13

PS: I think there is a general lack of URL encoding on all dynamic links, not only on document ID parameters.

PPS: This is my very first contribution to a GitHub open source project.. Please tell me if I did something wrong :)

### DIFF
--- a/views/collection.html
+++ b/views/collection.html
@@ -143,7 +143,7 @@ No documents found.
 {% if collectionName != 'system.indexes' %}
 <script type="text/javascript">
 function loadDocument(id) {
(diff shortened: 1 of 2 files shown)


andzdroid:
Thanks for the pull request!

PR #73, opened 2015-06-15
Allowing configurable bind addresses.

max-b:
Just seemed like a convenient option to be able to bind to arbitrary addresses. I might have overlooked some meaningful reason why this shouldn't be an option.

Thanks!


max-b:
Nevermind - realized that this is now an option in the develop branch, so I think this is a moot request. Sorry bout that...

PR #79, opened 2015-07-03
Expanded config and run instructions

andresgottlieb:
I think they were not completely clear for node beginners.


wulfsolter:
Thanks @andresgottlieb - not just beginners, but I know I wasted a few seconds figuring out I had to rename the config. A few seconds * many developers = lots of hours!

PR #98, opened 2015-09-25
Added GB and TB units.

jvilaplana:
When the input size is larger than 1 GB the user interface showed the value in Bytes. Now the user interface will also show the value in GB and TB before switching back to Bytes.


wulfsolter:
Thanks @jvilaplana

PR #100, opened 2015-09-25
Specify address to listen on in config.js

CraigKelly:
I prefer to listen only on localhost for safety.


wulfsolter:
Thanks @CraigKelly

PR #105, opened 2015-10-01
Support replica set

paton:
Apologies for the messy PR and lack of documentation. Hopefully will find time to clean this up soon (or if anyone wants to adapt this in to a better PR, definitely supportive of that!)

### DIFF
--- a/db.js
+++ b/db.js
@@ -14,8 +14,17 @@ var connect = function(config) {
     auto_reconnect: config.mongodb.autoReconnect,
     poolSize: config.mongodb.poolSize
   };
-  var db = new mongodb.Db('local', new mongodb.Server(host, port, dbOptions), {safe:true});
 
+  var db;
+
+  if (Array.isArray(host)) {
+    host = host.map(function(host) {
+      return new mongodb.Server(host, port, dbOptions);
+    });
+    db = new mongodb.Db('local', new mongodb.ReplSet(host), { safe: true, w: 0 });
+  } else {
+    db = new mongodb.Db('local', new mongodb.Server(host, port, dbOptions), {safe:true});
+  }
 
   var connections = {};
   var databases = [];
@@ -154,3 +163,4 @@ var connect = function(config) {
 };
 
 module.exports = connect;
+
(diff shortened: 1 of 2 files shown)


wulfsolter:
Nice work @paton 
Thank you

db.js
```diff
@@ -14,8 +14,17 @@ var connect = function(config) {
...
   };
-  var db = new mongodb.Db('local', new mongodb.Server(host, port, dbOptions), {safe:true});
 
+  var db;
+
+  if (Array.isArray(host)) {
+    host = host.map(function(host) {
+      return new mongodb.Server(host, port, dbOptions);
+    });
+    db = new mongodb.Db('local', new mongodb.ReplSet(host), { safe: true, w: 0 });
```
nodesocket: Don't you need to pass in the replica set name?

nodesocket:
Also, can we get documentation on how the `config.js` file looks using a replica set?

paton:
@nodesocket Regarding replica set name, I originally tried passing it in but was getting some errors. Removing it fixed that issue, but from what I've seen, the standard practice is to pass it in.

Re: docs, simply pass in an array of hosts to the "server" option instead of a single host. For example

mongodb: {
    server: ['mongoserver1.com', 'mongoserver2.com', 'mongoserver3.com'],
    port: 27000,
    ...
}

Unfortunately I'm jammed right now and don't have time to make these improvements, but if I (or someone) has time to refactor this, a much better approach than this PR would be to allow a mongodb connection string instead of passing in hostname, ports, RS name, etc. http://docs.mongodb.org/manual/reference/connection-string/

All of those variables are embedded in a mongo connection string.

ceefour:
@paton How do I do this from Docker image's ? `ME_CONFIG_MONGODB_SERVER` environment variable doesn't seem to support array.....

paton:
Hm, I'm not familiar enough with Docker or how env vars are handled in this repo.

@wulfsolter might know more.

knickers:
The problem will be with Linux environment variables. They don't support arrays, or at least I have never figured out how to do it.

ceefour:
@knickers we can use commas. or better yet just use connection string

PR #109, opened 2015-10-07
create image from official node repo

lholmquist:
The google/nodejs repo is deprecated and the docker build was broken because of it.

i updated the dockerfile to pull from the official node.js repo(the latest version)


wulfsolter:
Thanks

PR #121, opened 2015-10-29
Add SSL capabilities and configuration to mongo-express

iTerminate:
I've made changes that would add ssl capabilities only if user specifies it should be enabled. By default it is the same as running the application before.


wulfsolter:
Nice! Thanks @iTerminate!

PR #136, opened 2015-12-01
fixed bug of exporting: removed 'ISODate(' from exported JSON

fsubal:
Fixed the bug related to Issue #64. Broken format of `ISODate(` in exported json files is now removed.

Issue #64, opened 2015-05-20
ISODate format output bug


fbukevin:
I stored some crawled data in mongodb and there is a field "created" which value format looks like this in mongo-express：

![bug](https://cloud.githubusercontent.com/assets/4548874/7730273/e2fb7e9e-ff4b-11e4-9a41-26b2fcb868d7.JPG)

It looks fine, but when I tend to export it to JSON file, there appears a character error：
![export](https://cloud.githubusercontent.com/assets/4548874/7730303/1aacdea0-ff4c-11e4-90a1-9c98b5e10d31.JPG)

wulfsolter:
Closing as inactive

PR #137, opened 2015-12-03
fixed bug of exporting: removed 'ISODate(' from exported JSON

fsubal:
Same as #136, Retrying with another brunch.
Fixed the bug related to Issue #64.
Broken format of ISODate( in exported json files is now removed.


wulfsolter:
Awesome! Thanks heaps @fsubal :)

PR #140, opened 2015-12-08
Use the dockerfile from knickers/mongo-express

knickers:
This is mostly just to start the discussion. I'd like to make it as easy to merge with your deployment process as possible.

This builds the docker image from a specific npm release, so when you update the version number there, you would also update the version number here.

Running the node app behind *tini* makes the container gracefully stop, it wouldn't receive the sigterm otherwise.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,26 @@
-FROM node:latest
+FROM node:0.12-slim
 
-WORKDIR /app
-ADD . /app
+ENV TINI_VERSION 0.5.0
+RUN set -x \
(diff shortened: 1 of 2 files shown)


knickers:
Have you already set up a docker hub account? Mine is an automated build, so whenever the repo changes the image is rebuilt automatically. It's nice not having to manage *one more thing*.

Or if you want to just direct people to my image that's OK too :smiley:

knickers:
I changed the environment variable list to be a table. Let me know if you like it.

wulfsolter:
Wowsers! Thanks heaps @knickers :) Looks great!

PR #141, opened 2015-12-10
Add config.js

andresgottlieb:
mongo-express should come with a config.js file (exact copy of config.default.js). This way it will run out of the box for mongodb default installations, not needing to (first figure out, and then) make this copy manually. I strongly believe this friction is currently stopping a lot of people from using this tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,9 +66,7 @@ Or if you want to install a global copy:
 
 **To configure:**
 
-Copy or rename `node_modules/mongo-express/config.default.js` into a new file.
-
-Fill in your MongoDB connection details, and any other options you want to change.
+Modify `node_modules/mongo-express/config.js` to set your preferences, filling in your MongoDB connection details, and any other options you want to change.
 
 **To run:**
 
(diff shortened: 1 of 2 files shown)


wulfsolter:
While I agree in principle that a default state of "working out of the box" is nice, the proposed PR will break almost every current existing installation by overwriting the custom config.js everyone has created which is pretty annoying and unnecessary. Adding a second exact copy of config.default.js goes against DRY principles as well.

What if we modify the startup in app.js to something along the lines of:
* If a config.js exists, use that
* If valid CLI params are passed, use them.
* If invalid CLI params, show startup message
* Fallback to using config.default.js

andresgottlieb:
I agree with you and would happily participate of this. Shouldn't we separate the `if !config.js -> fallback to default` from the CLI params part? Or is there something I'm not seeing that makes them interdependant?

wulfsolter:
Ah, good call. So CLI > Custom Config > Default?

wulfsolter:
How about https://github.com/andzdroid/mongo-express/commit/18e326f24e35279465d7d205502c259fc643dd67 ? Uses CLI options over custom config.js, falling back to config.default.js

PR #143, opened 2015-12-10
add configuration for request size limit.

### LINKED ISSUE #142: entities too large, allowable entity size not configurable
this is an awesome tool. lots of mongo payloads are going to be larger than the `body-parser` default of 100KB.

a savvy user can reach in and change the code to update the limit, but there should be a piece of configuration that lets mongo-express users do it through a conventional interface.

pull request coming soon if I can keep my mind here in the same thread :)

nchase:
fixes #142

### DIFF
--- a/config.default.js
+++ b/config.default.js
@@ -56,6 +56,7 @@ module.exports = {
     cookieSecret: process.env.ME_CONFIG_SITE_COOKIESECRET || 'cookiesecret',
     sessionSecret: process.env.ME_CONFIG_SITE_SESSIONSECRET || 'sessionsecret',
     cookieKeyName: 'mongo-express',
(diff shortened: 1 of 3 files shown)


config.default.js
```diff
@@ -56,6 +56,7 @@ module.exports = {
     cookieSecret: process.env.ME_CONFIG_SITE_COOKIESECRET || 'cookiesecret',
     sessionSecret: process.env.ME_CONFIG_SITE_SESSIONSECRET || 'sessionsecret',
     cookieKeyName: 'mongo-express',
+    requestSizeLimit: process.env.ME_CONFIG_REQUEST_SIZE || '100kb',
```
nchase: I wonder: is there a better default size to use here for this project?

PR #144, opened 2015-12-12
Revert back to the original way of building docker

knickers:
The original way of building the docker image was to use the source from the current directory. That is better than the way I updated it to be, which is to pull a specific version from npm. (redundant because it's already here)

This also removes the need to have a second copy of the latest node package version number.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,17 +10,11 @@ RUN set -x \
 	&& tini -h \
 	&& apt-get purge --auto-remove -y ca-certificates curl
 
-ENV MONGO_EXPRESS 0.27.7
+WORKDIR /app
+COPY . /app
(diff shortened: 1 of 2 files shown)


knickers:
I've also updated the README to show how to build the container, because it seemed weird to point people to the knickers image.

p.s. Have you considered providing a ready-made image on [docker hub](https://hub.docker.com/)? Then we actually could point people to one that they don't have to build themselves.

wulfsolter:
Nice work :)

As for Docker Hub, I'd be more than happy if you set up a ready made image and then linked it in the docs? I've not used Docker Hub, but from what I can tell they're quite supportive of FOSS and offer free services for projects like this?

PR #148, opened 2015-12-17
Be able to see databases for MongoDB 3.2 

catorda:
The listDatabases changed to an array (looks like before it was an object with keys). This change would break backwards compatibility though. Should I check the version and change the way it iterates through the databases? I wasn't sure which mongo version it changed in though, maybe just checking for 3.x would be fine.

### DIFF
--- a/db.js
+++ b/db.js
@@ -154,13 +154,13 @@ var connect = function(config) {
   });
 
   return {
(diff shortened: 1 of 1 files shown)


wulfsolter:
Nice! Thanks! :)

PR #160, opened 2016-01-13
Bash can't pass literal 'true', just a truthy value

knickers:
I'm sorry, I should have tested this new environment variable better before making a pull request yesterday.


wulfsolter:
Nice catch :)

PR #162, opened 2016-01-14
Explicitly link the docker port to the host port

knickers:
I had always run it with a custom port link (`-p 8082:8081`) so I never tested it without one. Turns out it can't connect without it. Also docker connections don't like `localhost` sometimes, to I changed the docker default to `0.0.0.0`.

P.S. I like the new full width page layouts :smile:


wulfsolter:
Nice catch! Cheers man!

PR #163, opened 2016-01-15
docs: add SSL environment variables to README.md

micooz:
The application does support SSL, but I didn't find any usages about it until I checked out the code. So I added something to the README.


wulfsolter:
Thanks @micooz !

PR #171, opened 2016-02-02
Support MongoDB SSL connections.

ctigeek:


wulfsolter:
Nice! Thanks @ctigeek

PR #174, opened 2016-02-10
Regex search v2

tatakci:
Regex search v2


wulfsolter:
Thanks! Published to npm

PR #178, opened 2016-02-12
Make jsons in collection views collapsible, fixes #176

### LINKED ISSUE #176: Dynamic collapsing of subobjects
Hi, I just wanted to ask whether you could consider implementing collapsing subobjects in collection views. I have some subobjects that are pretty big and setting `maxPropSize` makes them collapsed completely - and when I click them, I see the whole huge subobject and have no way to hide it.
It'd be really cool if there was an option to dynamically (un)collapse the subobjects similar to [1]. I'd be glad to send a PR, but I'm no JS/NodeJS expert. Perhaps given some pointers, I could do that (like - is there actually a library for that? My googling doesn't reveal anything, but perhaps I'm just doing it wrong).

[1] http://jsfiddle.net/t5Hxk/5/

bkabrda:
This adds the ability to collapse subobjects using renderjson \[1\].
- renderjson is licensed under ISC \[2\], which according to wikipedia is functionally equivalent with MIT; hopefully that means that there should be no licensing issues, but IANAL
- The implementation is a bit naive - the whole page is loaded first and then a custom JS function traverses all `div.tableContent` elements and if one of them contains JSON string, it parses it and replaces it with renderjson result.
- I've added two configuration options for turning on/off using renderjson and for setting the default unfold level.
- The only issue that I'm experiencing is that the table columns change width when collapsing/uncollapsing. I'm not sure whether that's fixable - I'm not sure whether we should (can) fix it in the first place... I'm just mentioning it for your consideration.

I'm not an experienced JS/frontend developer, so if you see a better way to do something or if you would like me to change the behaviour in some way, please say so.

\[1\] https://github.com/caldwell/renderjson
\[2\] https://en.wikipedia.org/wiki/ISC_license

### DIFF
--- a/config.default.js
+++ b/config.default.js
@@ -125,6 +125,12 @@ module.exports = {
 
     //readOnly: if readOnly is true, components of writing are not visible.
     readOnly: false,
+
+    //collapsibleJSON: if set to true, jsons will be displayed collapsible
+    collapsibleJSON: true,
+    //collapsibleJSONDefaultUnfold: if collapsibleJSON is set to `true`, this defines default level
+    //  to which JSONs are displayed unfolded; use number or "all" to unfold all levels
+    collapsibleJSONDefaultUnfold: 1
   },
 
   // Specify the default keyname that should be picked from a document to display in collections list.
--- a/lib/routes/collection.js
+++ b/lib/routes/collection.js
@@ -184,6 +184,8 @@ var routes = function(config) {
           fields: jsonFields,
           defaultKey: defaultKey,
           edKey: edKey,
+          collapsibleJSON: config.options.collapsibleJSON,
+          collapsibleJSONDefaultUnfold: config.options.collapsibleJSONDefaultUnfold
         };
 
         res.render('collection', ctx);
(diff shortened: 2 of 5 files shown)


wulfsolter:
I've made two little fixes:

- was trying to call renderjson() on all things starting with `<` which included `<img ..` and `<video ..` so broke all sorts of client side JS. https://github.com/andzdroid/mongo-express/commit/52034fdd6d16494ceb5cff990f05f1a96644168a
- when expanding/collapsing, the newly appended '+', '-' links bubble up to trigger 'edit document' which gets quite annoying very quickly. Fix in mongo-express https://github.com/andzdroid/mongo-express/commit/a8bbfe774da761333f53f90b7941c7f9945a1ac0 and PR at renderjson https://github.com/caldwell/renderjson/pull/10

Thanks for your great work! :D

bkabrda:
Cool, thank you very much.
As for stopping the propagation, I actually had code for that [1] and it worked for me. I wonder why it didn't work for you... Probably you can remove that callback now, since you put it in renderjson.

[1] https://github.com/andzdroid/mongo-express/pull/178/files#diff-719c36e360c37fccfb9f4ed48a2bace6R219

wulfsolter:
It works the first time round on the root expand/collapse toggles... just not once you've expanded a hidden child that has toggles itself - toggles that weren't visible when page got created.

bkabrda:
Ah, thanks for the explanation!

wulfsolter:
And merged into upstream renderjson :) https://github.com/caldwell/renderjson/pull/10#issuecomment-183755289

PR #183, opened 2016-02-18
Add some environment variables for non-admin db connections

knickers:
I actually haven't tested this because I don't have a db with credentials :fearful:. Is there someone who can test these?


wulfsolter:
I don't either - I personally think it's a pretty solid call by Mongo to let auth be handled by OS and/or other services, rather than building clunky authentication in front of a DB.

Looks sensical, so peer-review :+1:

PR #205, opened 2016-03-21
removed all client-side references to {{ baseHref }} to force only relative urls

Morriz:
My first changes were hand picked and some necessary changes were missed. I now thoroughly hunted the last ones down.

### DIFF
--- a/lib/views/collection.html
+++ b/lib/views/collection.html
@@ -4,8 +4,8 @@
 
 
 {% block head %}
-<link href="{{ baseHref }}stylesheets/codemirror.css" rel="stylesheet">
-<script type="text/javascript" src="{{ baseHref }}javascripts/codemirror-compressed.js"></script>
+<link href="stylesheets/codemirror.css" rel="stylesheet">
+<script type="text/javascript" src="javascripts/codemirror-compressed.js"></script>
 <script type="text/javascript">
 'use strict';
 function getParameterByName(name) { name = name.replace(/[\[]/, "\\[").replace(/[\]]/, "\\]"); var regex = new RegExp("[\\?&]" + name + "=([^&#]*)"), results = regex.exec(location.search); return results === null ? "" : decodeURIComponent(results[1].replace(/\+/g, " ")); }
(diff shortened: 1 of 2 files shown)


knickers:
In my opinion, relative URLs should always begin with a slash `/` to reference the root. This is because without the leading slash, the relative URL is appended into the current URL.

So for example, if you are on `http://mongo-express.local/collection.html` and try to load the resource `stylesheets/codemirror.css`, your browser will look at the URL `http://mongo-express.local/collection.html/stylesheets/codemirror.css`.

But if the recourse's reltive URL is `/stylesheets/codemirror.css`, then the browser will look at `http://mongo-express.local/stylesheets/codemirror.css`

Morriz:
I do believe however that my changes make it work in both use cases :) I did test the changes, and so can you. Please do.

knickers:
You are correct @Morriz, everything seems to still work even without the leading slashes. I must be missing something. @wulfsolter, does the router or `express.js` do anything magical for things like this?

I can see that there are still lots of references to `baseHref` in the following files: `lib/views/database.html`, `lib/views/gridfs.html`, `lib/views/index.html`, and `lib/views/layout.html`. Are they also going to play a part in letting mongo-express run well in kubernetes?

Also, I'm seeing something else interesting. When I view the page source, all the resources loaded from `lib/views/layout.html` are rendering as relative URLs with a leading slash, which means `baseHref` is just a slash `/`. Is this supposed to happen or is it a bug?

From `lib/views/layout.html`:
```html
<link href="{{ baseHref }}stylesheets/bootstrap.min.css" rel="stylesheet">
<link href="{{ baseHref }}stylesheets/bootstrap-theme.min.css" rel="stylesheet">
<link href="{{ baseHref }}stylesheets/style.css" rel="stylesheet" />
```

Rendered output:
```html
<link href="/stylesheets/bootstrap.min.css" rel="stylesheet">
<link href="/stylesheets/bootstrap-theme.min.css" rel="stylesheet">
<link href="/stylesheets/style.css" rel="stylesheet" />
```

wulfsolter:
`baseHref` comes from `config.sites.baseUrl` https://github.com/mongo-express/mongo-express/blob/master/config.default.js#L80

knickers:
Oh, OK. So it is set to `/` in my config. Is it common for this to be set to the full base URL like `http://mongo-express.local/` or something? @Morriz, what is yours set to?

Morriz:
The original motive was to make it configurable, with slash as default I guess. But it's just not needed, nor wanted imo.

The css refs still need a hard link, as they are prefixed with the path of the html file.

Morriz:
Just check out my config, nothing there. Works smooth now

Issue #241, opened 2016-07-12
Replace line 479 of lib/view/collection.html or you will not be able to delete a collection


manuel-di-iorio:
Replace:

    var $parentForm = $('#' + target.attr('childof'));
with:

    var $parentForm = $('#' + $target.attr('childof'));

The error is thrown when you click on the delete button, inside the collection page itself

    posts:658 Uncaught ReferenceError: target is not defined

PR #248, opened 2016-08-13
js variable name fix

muxahuk:
should fix https://github.com/mongo-express/mongo-express/issues/241


knickers:
Thanks @muxahuk! :laughing:

PR #249, opened 2016-08-13
add gridFSEnabled option in configs.default file

muxahuk:
So that users could see where to change these param


wulfsolter:
Thanks!

PR #259, opened 2016-09-13
Fixing the mountpath based on request.

### LINKED ISSUE #256: Incorrectly defined mountpath when mongo-express used as middleware
Here is my setup: 
```
var mongo_express = require('mongo-express/lib/middleware');
var mongo_express_config = require('mongo-express/config.default')

app.use('/mongo_express', mongo_express(mongo_express_config))
```

in `config.default` (for simplicity) I set
```
baseUrl: process.env.ME_CONFIG_SITE_BASEURL || '/mongo_express/'
```
When I open the url in my browser, static assets are not loaded correctly: they are not prefixed with `/mongo_express/`, and so not found. That is, here is the path with which I can access an image - `http://localhost:1337/mongo_express/images/mongo-express-logo.png`, but what the page requests is `http://localhost:1337/images/mongo-express-logo.png`, which is not found. 
Here is the part of the code of the main page:
  ```
<base href="/">
  <meta charset="utf-8">
  <title>Home - Mongo Express</title>
  <meta name="viewport" content="width=device-width, initial-scale=1.0">

  <link href="/stylesheets/bootstrap.min.css" rel="stylesheet">
  <link href="/stylesheets/bootstrap-theme.min.css" rel="stylesheet">
  <link href="/stylesheets/style.css" rel="stylesheet" />
```

We see that the paths are not prefixed. 
When I mount the app separately with `node app --port 3000` everything works just fine.

I traced the origins of the problem back to the line 
`res.locals.baseHref       = req.app.mountpath + (req.app.mountpath[req.app.mountpath.length - 1] === '/' ? '' : '/');` in `lib/router.js`
where the `req.app.mountpath` is equal to '/' instead of correct mountpath ('/mongo_express/'). So it should be easy to fix.

Thanks!

levigroker:
Performing the same mechanism parse-dashboard does to handle being installed as middleware. This fixes #256, in my testing.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -61,7 +61,9 @@ var router = function (config) {
   // view helper, sets local variables used in templates
   appRouter.all('*', function (req, res, next) {
     // ensure a trailing slash on the baseHref (used as a prefix in routes and views)
-    res.locals.baseHref       = req.app.mountpath + (req.app.mountpath[req.app.mountpath.length - 1] === '/' ? '' : '/');
+	var mountPathLength       = req.originalUrl.length - req.url.length;
+    var mountPath             = req.originalUrl.slice(0, mountPathLength);
+    res.locals.baseHref       = mountPath + (mountPath[mountPath.length - 1] === '/' ? '' : '/');
     res.locals.databases      = mongo.databases;
     res.locals.collections    = mongo.collections;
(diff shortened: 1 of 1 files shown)


levigroker:
Thanks!

PR #262, opened 2016-09-24
define mongodb host and port on CLI

PascalPflaum:
Allowing the user to define mongoDB Host and Port as argument for the CLI


wulfsolter:
Thanks for writing docs too @PascalPflaum! :)

PR #263, opened 2016-09-26
Refreshing collections when viewing database

gvillenave:
I noticed that if new collections are added after launching the app, they actually never appear on the interface because we never refresh the list. This quick fix forces refreshing the collections list when accessing the page.


wulfsolter:
Thanks @gvillenave!

Issue #272, opened 2016-10-18
Can't connect to replica set


jamrizzi:
I get the following error when I try and start mongo-express.

```
/home/jamrizzi/GroupThreads/mongui/node_modules/mongodb/lib/replset.js:360
          process.nextTick(function() { throw err; })
                                        ^
MongoError: no valid replicaset members found
    at Timeout.<anonymous> (/home/jamrizzi/GroupThreads/mongui/node_modules/mongodb-core/lib/topologies/replset.js:645:33)
    at tryOnTimeout (timers.js:232:11)
    at Timer.listOnTimeout (timers.js:202:5)
```

I have it configured as a replica set, and I have verified that all the information I entered is correct by trying it with the mongodb CLI.  Below is the way I have it configured.

```
  mongo = {
    db:       'admin',
    host:     [
        'rest-api-data-shard-00-00-xemv3.mongodb.net',
        'rest-api-data-shard-00-01-xemv3.mongodb.net',
        'rest-api-data-shard-00-02-xemv3.mongodb.net'
    ],
    password: 'myspecialpassword',
    port:     27017,
    ssl:      true,
    url:      'mongodb://rest-api-data-shard-00-00-xemv3.mongodb.net:27017,rest-api-data-shard-00-01-xemv3.mongodb.net:27017,rest-api-data-shard-00-02-xemv3.mongodb.net:27017/admin?replicaSet=rest-api-data-shard-0',
    username: 'myspecialusername'
  };

```

dozoisch:
I wonder if it's because the replicaSet name is not set properly using an array of hostnames. Because basically it seems like mongo.host is used but not mongo.url. Thus the `replicaSet=rest-api-data-shard-0` is never "lost" in the process.

jamrizzi:
I just verified that is not the issue.

```
      db = new mongodb.Db('local', new mongodb.ReplSet(host, {
          poolSize:    config.mongodb.poolSize,
          ssl:             config.mongodb.ssl,
          sslValidate: config.mongodb.sslValidate,
          sslCA:         config.mongodb.sslCA,
          replicaSet:  'rest-api-data-shard-0'
      }), {
          safe: true,
          w: 0,
          authSource: 'admin'
      });
```

dozoisch:
@jamrizzi That snippet in your last post works or not?

jamrizzi:
It does not work. I'm still getting the same error.

dozoisch:
[...]

PR #274, opened 2016-10-18
added support for mongo url driver

jamrizzi:
This should supposedly work.  However, it doesn't quite work yet.  It probably needs a bit of tweaking.  I did however do the hard work of getting the asynchronous MongoClient.connect() function loaded in a way synchronous way that should work with the rest of the db.js.

### DIFF
--- a/config.default.js
+++ b/config.default.js
@@ -23,7 +23,7 @@ if (process.env.VCAP_SERVICES) {
     password: 'pass',
     port:     27017,
     ssl:      false,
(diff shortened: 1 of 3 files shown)


GreenPioneer:
I know its only been two days but were do we stand on this PR ?

I really like it and could really use it

dozoisch:
@GreenPioneer this was just a WIP and is not currently working. It was an attempt for #272. We're open to PRs.

dozoisch:
I closed it for now. If we need it I will reopen.

PR #277, opened 2016-10-26
allow 24 digit hex string in ObjectID query

### LINKED ISSUE #251: Make the ObjectID(...) wrapper optional
When searching by ```_id``` it would be easier if we don't have to use the ObjectID(...) wrapper since the type (String, ObjectID...) was already choose.

heartnetkung:
fix #251 
So we can do something like this.
screenshot from 2016-10-26 16 28 01

### DIFF
--- a/lib/bson.js
+++ b/lib/bson.js
@@ -100,6 +100,14 @@ exports.toObjectId = function (string) {
   }
 };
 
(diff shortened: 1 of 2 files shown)


dozoisch:
Awesome thanks!

PR #281, opened 2016-10-28
Implement index management

heartnetkung:
## Pull Request Overview
- remove index listing in collection stats
- add "reindex" button in the tools section (make sure you have mongodb permission to do so)
- add new index listing section
  - it shows name, keys, index size, special attributes, and a button to drop index
  - the special attributes are all values given by db.collection.indexes except name, v, key, and ns 

## Example Screenshot
screenshot from 2016-10-28 18 41 02


dozoisch:
That looks really good! I'll try it a bit later today and merge :)!

PR #282, opened 2016-11-01
Fix typo in config.default.js

SimonLammer:


SimonLammer:
Wow. That is as fast as it can get! (not even a minute)

PR #289, opened 2016-11-28
feature: delete all documents matching a query

xavierpriour:
Hi,

I needed the ability to delete a hundred documents returned by a query. Not the whole collection, but too much to do it one by one.

So this PR adds a "Delete" button under the query form of the collections view that lets users delete the documents based on the query they input.

From a technical point of view, I did:

- extract the query building part of viewCollection so that both methods use the same code.
- hook into DELETE /db/collection and deletes if there is a query, drops if there isn't
- add a 'count' property to the rendering contrext, with the total number of documents returned by the query

Let me know if you need anything else to merge it.

### DIFF
--- a/lib/views/collection.html
+++ b/lib/views/collection.html
@@ -159,6 +159,18 @@ jQuery(document).ready(function ($) {
       </form>
     </div>
   </div>
+  {% if !settings.read_only && count > 0 %}
+  <p>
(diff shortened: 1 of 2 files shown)


dozoisch:
Hi! thanks a lot for this PR! It looks good, but I'll test it locally today and will merge afterwards!

lib/views/collection.html
```diff
@@ -511,6 +523,15 @@
     }
   );
 
+  $('#deleteListConfirmButton').on('click', function (e) {
+    console.log('deleteList', e);
```
dozoisch: whoop! seems like you forgot a console.log

xavierpriour: You got it!

PR #292, opened 2016-12-01
Don't enter in edit mode with selection

RrOoSsSsOo:
Into Collection page are listed all documents, if there is a mouse selection (for copy-paste a string) then mongo-express don't enter in Edit Mode for Document.


jasstionzyf:
@fracmak  @RrOoSsSsOo  any progress to merge this feature? 
currently it is very not convenient to copy content

PR #294, opened 2016-12-01
Remain in Edit Mode on update Document

RrOoSsSsOo:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/294)
---

Remain in Edit Mode page of JSON document after click "Save" on Edit page

### DIFF
--- a/lib/routes/document.js
+++ b/lib/routes/document.js
@@ -92,7 +92,9 @@ var routes = function (config) {
       }
 
       req.session.success = 'Document updated!';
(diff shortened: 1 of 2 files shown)


shakaran:
> Should we use a boolean env var to enable/disable the Edit Mode behavior?

Yes, give options to the user it is always a nice way

rtritto:
@BlackthornYugen replaced by #910, can be closed

Issue #297, opened 2016-12-22
I can delete indexes even though options.readOnly = true


hyeonsoo:
I set **options.readOnly = true** in my config.js file.

Other components of writing were removed from the page as expected.

But I can still see the delete buttons on indexes section.

![image](https://cloud.githubusercontent.com/assets/15084045/21413484/ecfb142c-c83a-11e6-8e00-dcd06262f2da.png)

Please fix this.

Thank you.

dozoisch:
Good one, will try to do it tomorrow unless someone volunteers!

hyeonsoo:
Even though you hide buttons from the page,
What if someone manually type the URL of dropping index?
As I see the router logic, I found there's no checking of options.readOnly at all.
So anyone can even do any operations regardless of readOnly option, right?

dozoisch:
0.34

PR #298, opened 2016-12-22
Fix issues #297 deleting indexes even though options.readOnly = true

jovehammer16:
#297 
update collection.js to check config.options.readOnly before dropIndex
update collection.html to hide the delete button ready only.


dozoisch:
Awesome! @jovehammer16 I'll test it locally tonight and will merge and release a new version!

PR #304, opened 2017-02-01
Fix content-disposition for non-ascii header, set config.gridFSEnabled via env vars

ivliag:
1. Non-ascii symbols aren't allowed in setHeaders in Node 5+
So, if I have cyrillic filename for example, it will brake everything.
Similar issue: https://github.com/nodejs/node/issues/1693

2. It's very useful to be able to set config.gridFSEnabled via env vars, when using Docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Fill in your MongoDB connection details and any other options you want to change
 If you installed it globally, you can immediately start mongo-express like this:
 
     mongo-express -u user -p password -d database
-    
+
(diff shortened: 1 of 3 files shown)


wulfsolter:
Looks good @ivliag :)

Can you please add documentation for config setting to your PR?

ivliag:
@wulfsolter yeah, sure, fixed it

wulfsolter:
Nice one! Thanks @ivliag

Issue #303, opened 2017-01-27
Collections with / (forward slash) in the name can't be viewed


marcp:
The collection names in a database that we've inherited have names like /group/section/key.  They appear to look fine at the top level when viewing the database, but when I click on one of them I get:

Cannot GET /db/mydatabase//group/section/key

I just want to know if slashes in collection names are unsupported or if I'm doing something wrong in my config such that it's not working.

Thank you!

dozoisch:
Hello! 

I don't think we've had that case before, so I don't think it's a configuration issue! We're open to PRs if you want to try to add support, else it will be when time permits.

wulfsolter:
Currently slashes are unsupported. 

It would be a **massive** undertaking to add this functionality, as slashes are used for database/collection/document heirachy. Mongo-Express does not cover every possible feature, and for most people (who use sane names) the slash separation results in meaningful and easily comprehensible URLs.

While we know your current collection names are "valid" [according to spec](https://jira.mongodb.org/browse/SERVER-3229), there is still no way I [would name my child Null](https://www.wired.com/2015/11/null/)  just because it's going to lead to problems *somewhere*. Similarly I'm never going to make a Postgres table called "' OR 1=1; DROP TABLE USERS;" because sooner or later something will happen.

marcp:
Thank you for the responses.

dozoisch:
Actually, I will reopen this issue as I think we can fix it by just url encoding the collection name everywhere. I'm working on it!

marcp:
Thank you for doing this!  I just tried it out and it works great for us.  Greatly appreciated.

dozoisch:
My pleasure! I'll cut a real release soon, but there are a few more things I want to finish first.

PR #309, opened 2017-02-13
Added support for / in collection names

dozoisch:
Closes: #303

@marcp @wulfsolter what do you think about that? It's a bit ugly with the url_encode everywhere, but it seems to work just fine.

image


wulfsolter:
Very nicely done :D

PR #312, opened 2017-02-19
Changed the way scripts are loaded into page

dozoisch:
- use node_modules to load scripts. This makes it a lot easier to update
them
- use node_modules for css files. (Take them from distributions and copy
them in build directory. This includes all codemirror themese
- removed prettify (doesn't seem to be used anywhwere)
- removed all assets that are now in webpack build

- added webpack to build scripts
  - each page has his own bundle
  - two common: codemirror & vendors.
    - vendors has jquery & bootstrap
    - codemirror has codemirror!
  - jquery, bootstrap & codemirror are now node_modules

- build is done on install for production.
- npm run start-dev builds & runs for staging
- npm start now start in production

---

Would love to get some feedback on this! @wulfsolter @heartnetkung


dozoisch:
@wulfsolter trying to :)! Next step is tests for views. Want to make sure that we can move fast without breaking things

PR #315, opened 2017-02-20
Added test for index of db, collection and document

dozoisch:
- fixed missing closing tag in database.html template

---

Would love to get some feedback about the structure of tests. (see also #314)


dozoisch:
Just need to solve the race condition issue with the collection list... and good to go!

PR #320, opened 2017-02-22
[Fixes #319] Fix router baseHref

### LINKED ISSUE #319: Middleware mountpath isn't correct if middleware not mounted at /
Alright, almost everything's working. The only issue I now have is where the baseHref value trims off the last character in the URL. I've traced the issue back to [here](https://github.com/mongo-express/mongo-express/blob/master/lib/router.js#L64)

In my case, the middleware is mounted at some location, say `dbs/bar`. The `req.originalUrl` in this case is `dbs/bar`, however,  `req.url` in the router is `/`. This results in `mountPath` being `dbs/ba`, and everything gets kind of messed up.

I'll happily put in a PR for this, but I wanted to check with you first on best course of action. Maybe just remove the trailing `/` from `req.url`?

MikeyBurkman:
Hopefully this works. I tested both running from the command line, and running in my app which uses mongo-express as middleware routed to various urls. If you have a more in-depth regression test process, I would definitely try that :)

To solve it, I started out by printing the `req.originalUrl` and `req.url` while navigating through the app, and looking at how to transform those values into a base url that would work for all of them. My test data essentially looked like this:

```js
buildBaseHref('/myRoutes/foo', '/');
buildBaseHref('/myRoutes/foo/db/FH_LOCAL/', '/db/FH_LOCAL/');
buildBaseHref('/myRoutes/foo/db/FH_LOCAL/public/css/bootstrap-theme.min.css', '/db/FH_LOCAL/public/css/bootstrap-theme.min.css');
buildBaseHref('/myRoutes/foo/db/FH_LOCAL/public/img/gears.gif/db/FH_LOCAL', '/db/FH_LOCAL/public/img/gears.gif/db/FH_LOCAL');
```

In every case, the returned value should be `/myRoutes/foo/`, which is what the buildBaseHref() function does, by removing the req.url from the end of the req.originalUrl. (Unless it's the base route, which is unfortunately a special case.) Hooray for TDD.

Let me know if I missed something. I likely did.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -58,12 +58,23 @@ var router = function (config) {
     appRouter.use(errorHandler());
   }
 
+  var buildBaseHref = function (originalUrl, reqUrl) {
+    var addTrailingSlash = function (s) {
+      return s + (s[s.length - 1] === '/' ? '' : '/');
+    };
+
+    if (reqUrl === '/') {
+      return addTrailingSlash(originalUrl);
+    }
+
+    var idx = originalUrl.lastIndexOf(reqUrl);
+    var root = originalUrl.substring(0, idx);
+    return addTrailingSlash(root);
+  };
+
   // view helper, sets local variables used in templates
   appRouter.all('*', function (req, res, next) {
-    // ensure a trailing slash on the baseHref (used as a prefix in routes and views)
-    var mountPathLength       = req.originalUrl.length - req.url.length;
-    var mountPath             = req.originalUrl.slice(0, mountPathLength);
-    res.locals.baseHref       = mountPath + (mountPath[mountPath.length - 1] === '/' ? '' : '/');
+    res.locals.baseHref       = buildBaseHref(req.originalUrl, req.url);
     res.locals.databases      = mongo.databases;
     res.locals.collections    = mongo.collections;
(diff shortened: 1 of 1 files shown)


dozoisch:
Wooh,  Thanks!

Looks good. I'll test it tonight.  If everything works, I'll go ahead and merge it.

Small nit: You are missing a `;` on line 64 (this is why build fails!)

MikeyBurkman:
Ah whoops! Funny story -- it actually failed to build locally, but because I'm getting this bizarre error:
```
  1) Router collection GET /db/<dbName>/<collection> should return html:
     Error: expected 200 "OK", got 302 "Found"
```
I'm not totally sure why supertest would ever return a 302. Any ideas? (It actually failed the same way when I attempted a completely unrelated change earlier today.)

dozoisch:
So there is a redirect when collection name is empty or not found (instead of a 404) currently. 

I had those issues a lot and it was due to a race condition in tests. Which I patched with a timeout while waiting for a real fix. (I want to refactor the whole connection module)

dozoisch:
Haven't had time to test it yet, it's on my todo list. Sorry for delay

PR #326, opened 2017-03-25
Add CreateIndex Function

DarrenWong:
Add the function button after the Collection-> New document but a little confuse how to add the textarea effect and validation(seems not working when add into the lib/scripts/collection.js).Hope can help to improve
image


dozoisch:
Hey thanks a lot! I'll try it out over the weekend and see if I can make it work

DarrenWong:
Thank you for your kindness. 
And this morning I found that it 's stupid thinking I just running the code after changing like i write GUI previous. It need to take effect after npm install. Sorry that I am a fresher in NodeJs. Thanks for further improvement. 
Creating Index in MongoDB speed up the query many times but it may take longer time and blocking others activities when running. 
I may add some options can choose to run it background later
Builds the index in the background so that building an index does not block other database activities.

PR #339, opened 2017-04-30
Changed Database Name Validation Process

### LINKED ISSUE #327: DB and collection name are too restrictive
I tried a name with a dash and it doesn't work as the message says : "Database names must begin with a letter or underscore, and can contain only letters, numbers, underscores and dots."

From the doc, it's allowed to use a dash : https://docs.mongodb.com/manual/reference/limits/#naming-restrictions

dozoisch:
- closes #327 

@abellion @MikeyBurkman What do you think of this?

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -105,3 +105,16 @@ exports.roughSizeOfObject = function (object) {
 exports.buildCollectionURL = function (base, dbName, collectionName) {
   return base + 'db/' + dbName + '/' + encodeURIComponent(collectionName);
 };
+
+exports.isValidDatabaseName = function (name) {
+  if (!name || name.length === 0 || name.length > 63) {
+    return false;
+  }
+
+  // https://docs.mongodb.com/manual/reference/limits/#naming-restrictions
+  if (name.match(/[/. "$*<>:|?]/)) {
+    return false;
+  }
+
+  return true;
+};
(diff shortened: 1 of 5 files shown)


lib/utils.js
```diff
@@ -105,3 +105,16 @@ exports.roughSizeOfObject = function (object) {
 exports.buildCollectionURL = function (base, dbName, collectionName) {
   return base + 'db/' + dbName + '/' + encodeURIComponent(collectionName);
 };
+
+exports.isValidDatabaseName = function (name) {
+  if (!name || name.length === 0 || name.length > 63) {
```
MikeyBurkman: Technically, you don't really need the name.length === 0 check, as an empty string is falsy.

lib/utils.js
```diff
@@ -105,3 +105,16 @@ exports.roughSizeOfObject = function (object) {
...
+    return false;
+  }
+
+  // https://docs.mongodb.com/manual/reference/limits/#naming-restrictions
+  if (name.match(/[/. "$*<>:|?]/)) {
```
MikeyBurkman: Oddly enough, the docs don't say anything about single quotes. Weird.

lib/utils.js
dozoisch: Oh yeah, I'll remove that check.

PR #345, opened 2017-05-30
Use stream for exporting collections

ahmedissa:
User can't export big collections.

Cause: toArray function loads all documents in memory. If server ram didn't fit, mongo-express will exist.


MikeyBurkman:
I haven't had a chance to actually test this, but the changes look solid to me

ahmedissa:
No problem, glad to help.

Issue #257, opened 2016-09-09
csrf issue when setting ME_CONFIG_SITE_BASEURL


davvdg:
Hi

I'm trying to use mongo express through a mesos/marathon setup provided by mesosphere dcos. The web apps exposed through this software are in a subfolder ( http://<dcos_master_ip>/service/<servicename>/..., servicename is mongoexpress in this case)
So I've tried to set the env variable  ME_CONFIG_SITE_BASEURL=/service/mongoexpress/ but this make the csrf module to fail, with a 502 error:

```
Error: misconfigured csrf
   at csrf (/node_modules/csurf/index.js:71:19)
   at Layer.handle [as handle_request] (/node_modules/express/lib/router/layer.js:95:5)
   at trim_prefix (/node_modules/express/lib/router/index.js:312:13)
   at /node_modules/express/lib/router/index.js:280:7
   at Function.process_params (/node_modules/express/lib/router/index.js:330:12)
   at next (/node_modules/express/lib/router/index.js:271:10)
   at expressInit (/node_modules/express/lib/middleware/init.js:33:5)
   at Layer.handle [as handle_request] (/node_modules/express/lib/router/layer.js:95:5)
   at trim_prefix (/node_modules/express/lib/router/index.js:312:13)
   at /node_modules/express/lib/router/index.js:280:7
   at Function.process_params (/node_modules/express/lib/router/index.js:330:12)
   at next (/node_modules/express/lib/router/index.js:271:10)
   at query (/node_modules/express/lib/middleware/query.js:49:5)
   at Layer.handle [as handle_request] (/node_modules/express/lib/router/layer.js:95:5)
   at trim_prefix (/node_modules/express/lib/router/index.js:312:13)
   at /node_modules/express/lib/router/index.js:280:7
```

removing the env variable allows mango express to show up, but without all the static files, scripts and so on, making it unusable.

trivender:
similar issue.. please help...

albanm:
Hello,

Should this really be closed ? I have the same error with the latest docker image. For example with this docker-compose instruction : 

    mongoexpress:
      image: mongo-express:0.38.0
      environment:
        - ME_CONFIG_SITE_BASEURL=/mongo-express/

The stack :

```
Error: misconfigured csrf
   at csrf (/node_modules/csurf/index.js:71:19)
   ...
```

jbruggem:
I have the same error as well.

KrishnaPuppala:
[...]

PR #347, opened 2017-06-29
Solution for #257

margau:
Should be the solution for CSRF not working when Base URL is different. (Issue #257 )


dozoisch:
Thanks! I'll try it tonight and merge.

dozoisch:
@margau Sorry about delay! Merged and released as 0.41

PR #350, opened 2017-07-19
Support for connection strings so this works easily with atlas

spacem:
For  #272 

Added environment variable ME_CONFIG_MONGODB_URL, also --url command line parameter and connectionString option to the configuration object.

When connection string is specified other connection options (like ssl, etc) are ignored. In theory these could still be used but when I was doing this I still got errors from atlas about the replica set.

Since you can specify a database in the connection string. I connect to this database if the configuration does not specify a limited set of databases.


dozoisch:
Thanks! WIll be looking at it tonight

dozoisch:
Haven't had time yesterday, it's on top of my to do list.

Issue #367, opened 2017-09-25
Accessing documents with BinData as value for `_id`


riker09:
Hello there. I'm new to MongoDB and C#, so please don't frown when this is a stupid question. But I'm unable to access documents in the web client where the value for the `_id` column (property?) was generated from a C# `Guid`.

Here's my (simplified) C# class:

```csharp
    public class MyClass
    {
        [BsonId]
        public Guid MyId {get; set; }

        public string MyPropertyA { get; set; }
    }
```

Now when I store a document:

```csharp
var myClassObj = new MyClass
{
    MyId = Guid.NewGuid(), // Let's assume it is "C56A4180-65AA-42EC-A945-5FD21DEC0538"
    MyPropertyA = "Foo"
}
await mongoClient.SaveDocument(myClassObj.ToBsonDocument());
```

This is what ends up in the mongo database (retrieved with mongo cli):
```javascript
{ "_id" : BinData(3,"gEFqxapl7EKpRV/SHewFOA=="), "MyPropertyA": "Foo" }
```

This is what is rendered in the collection view:

![mongodb guid error](https://user-images.githubusercontent.com/568497/30693054-3512ab8e-9ece-11e7-9ad9-bc4aa78304ac.PNG)

And I cannot open the document. The error message is `Document not found!`.

I'm not sure if this is an error with mongo, mongo-express or C# Guids. I'm still learning and would appreciate some insights.

> Issue was first (mistakingly) reported here: https://github.com/mongo-express/mongo-express-docker/issues/5

heartnetkung:
Mongo-express does not support binary id. Though what you're doing is a little weird. 

Normally, the _id field is either number, string, or MongoID. Use number or string if you really need to specify your primary unique key (such as when you import your data from elsewhere), and use default MongoID otherwise (insert your data without _id field).

Hope it helps.

riker09:
So what you're saying is that my data has a Guid ID and I cannot use that as a primary key, is that it? Instead I should let Mongo choose an ID by itself and store my ID in another field. And if I ever need to retrieve my data by (my) primary key just do a search on my ID field? Okay, will do so.

It just feels a little bit weird to me, as well. Mongo seems to be just fine with a Binary ID as PK. Judging from my point of view it seems like missing feature in mongo-express. But I can live with that since mongo-express is supposed to be a debugging tool after all. :slightly_smiling_face:

dozoisch:
[...]

PR #368, opened 2017-10-05
Handle Binary data in primary key

riker09:
As discussed in issue #367 this PR addresses the issue that documents with a Binary as primary key (as opposed to ObjectID) are not accessible with mongo-express. Please note that there is certainly room for improvement here. But it works on my machine. :see_no_evil:


dozoisch:
Alright sorry about the delay, looking at it right now!

riker09:
No problem, I just got some notifications on merged pulled requests from April 2015 on another project.

PR #382, opened 2017-12-15
Prevent uncaught error if collection has been deleted.

mattcollier:
If one is at the 'Viewing Database' screen for a database with collections where collections are displayed.  If those collections are dropped (using the CLI), then one attempts to click on a collection (without first doing a refresh), the node process fails due to an uncaught error at: https://github.com/mongo-express/mongo-express/blob/master/lib/routes/collection.js#L231

`TypeError: Cannot read property 'indexSizes' of undefined`

### DIFF
--- a/lib/routes/collection.js
+++ b/lib/routes/collection.js
@@ -154,6 +154,10 @@ var routes = function (config) {
 
     req.collection.find(query, projection, query_options).toArray(function (err, items) {
       req.collection.stats(function (err, stats) {
+        if (stats === undefined) {
+          req.session.error = 'Collection not found!';
+          return res.redirect('back');
+        }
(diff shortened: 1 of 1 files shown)


dozoisch:
Thanks! I'll cut a release tonight

mattcollier:
@dozoisch you feel like this check is really in the right place?  I felt like it was a bit of a hack, but at least a place to start from.

MikeyBurkman:
@mattcollier I agree. If stats returns any error, it'll give the "collection not found" error message, which might be disguising the real error.

dozoisch:
@MikeyBurkman can probably add a log of the error in the terminal.

Issue #385, opened 2018-01-12
export query result


scne:
Is it possible to export query result on collection?

wulfsolter:
PRs are very welcome :) Should be pretty straightforward to implement.

MikeyBurkman:
This was implemented in #388, will be in the next release.

aogilvie:
Is this implemented for export .csv?
When ever I run a query and hit export .csv it exports the entire collection :(

mongo 4.0.1
docker image latest

gozn:
any update? latest version still can't export

rtritto:
@gozn this is working on the v1.1.0-rc-3

PR #388, opened 2018-01-27
export query results from collections

schuttsm:
Addresses: https://github.com/mongo-express/mongo-express/issues/385

* add export collection query results to json button
* change export collection api call to filter based on query

### DIFF
--- a/lib/routes/collection.js
+++ b/lib/routes/collection.js
@@ -118,18 +118,6 @@ var routes = function (config) {
     return query_options;
   };
 
-  exp._getQuery = function (req, res) {
-    var query;
-    try {
-      // if no criteria were passed, return whole list
-      query = exp._buildMongoQuery(req, res) || {};
-    } catch (err) {
-      req.session.error = err.message;
-      return res.redirect('back');
-    }
-    return query;
-  };
-
   exp._getProjection = function (req) {
     var projection      = {};
     var jsonProjection  = req.query.projection  || '';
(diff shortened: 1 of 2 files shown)


lib/routes/collection.js
```diff
@@ -119,7 +115,10 @@ var routes = function (config) {
       limit:  limit,
       skip:   skip,
     };
+    return query_options;
+  };
 
+  exp._getQuery = function (req, res) {
```
MikeyBurkman: I like having these new functions, but the error case when parsing the query in this function might cause issues. Before, the `return` in the case of a bad query would stop execution of the viewCollection() function, but now it'll just stop execution of this new function, and `query` will be some nonsense value on line 147. The session will redirect, but the query will still run (maybe) and who know what happens after that.

Might have to just let the error propagate up, surround line 147 with a try/catch, and then handle the error and redirect there.

(That being said, though it's almost certainly out of scope here, I'd love for the app to restore the query/projection values when there's an error parsing the query.)

schuttsm: @MikeyBurkman - good catch!

I moved the try catch blocks up a level so the parent functions are returned on error.

MikeyBurkman:
Alright, so I ran this locally and it worked fine. The only maybe-issue I see is that it exports only as the standard format, while I personally was expecting JSON array. It's of course not wrong (my intuition was wrong), but it makes me wonder if we can make this even better.

I feel like the next logical step after this would be to allow exporting the current query in the other formats. @dozoisch What do you think? Instead of adding a new export button, maybe just make the current export buttons export what you've queried? It's a change in current behavior, but I feel it's more intuitive that way.

I think the hard work is already done--we would just need to use the same query/project/query_options in the `collection.find()` calls in `exportColArray` and `exportCsv`, and of course pass in those options to each button via the html, like you're already doing.

schuttsm:
@MikeyBurkman - thank you for the feedback.

I removed the new export button & hooked the query filter & projection inputs into the existing export buttons and you were right - so much more intuitive.  How does it look now?

MikeyBurkman:
@schuttsm Thanks!

Issue #380, opened 2017-12-05
Cannot create collection name with dash in it


rayfoss:
A name like `student-progress` won't be accepted in 0.44 despite the mongodb handling it.

![screen shot 2017-12-05 at 5 21 11 pm](https://user-images.githubusercontent.com/876076/33636193-cd5d68fa-d9e0-11e7-936a-9a4e94f4e986.png)

MongoDb version: 3.4.10 (Atlas)
Node: v8.9.0

dozoisch:
The tooltip message seems outdated. Are you sure you are running 0.44?

![image](https://user-images.githubusercontent.com/4422516/33636408-d2f592d4-d9d0-11e7-9275-0929a30e5714.png)

This is what I would be expecting

dozoisch:
The message showing in your screenshot is from before the fix. (As seen in this commit: https://github.com/mongo-express/mongo-express/pull/339/files#diff-c944be67c1e215d32e8b0c3ea266aa56L5)

dozoisch:
OH collections, that is a different issue... I can look into it over the weekend. Or we accept prs :)

rayfoss:
@dozoisch glad I wasn't chasing ghosts... I missed that too. I would have PR'ed first, but I'm not familiar with this project... I couldn't at a glance see how the request is being made... let alone how your validation werks. Will checkout it out later, awesome proj btw

PR #389, opened 2018-01-28
allow dashes in collection names

schuttsm:
Addresses https://github.com/mongo-express/mongo-express/issues/380

* allow dashes ("-") in the collection names


schuttsm:
@MikeyBurkman done

PR #395, opened 2018-03-12
Fix #391 by supporting NaN, +Inifity, -Infinity values

### LINKED ISSUE #391: NaN and Infinity are displayed as null in document edition
Through the mongo-express web interface when editing a document stored in mongo I get the display below:
```javascript
{
    "_id": "test",
    "inf": null,
    "minf": null,
    "nan": null,
    "array": [
       12.05,
        null,
        458,
        null,
        14523,
        12.4538,
        null
    ]
}
```
While the real object stored in database is containing:
```javascript
{
    "_id": "test",
    "inf": Infinity,
    "minf": -Infinity,
    "nan": NaN,
    "array": [
       12.05,
        NaN,
        458,
        Infinity,
        14523,
        12.4538,
        -Infinity
    ]
}
```
I know the second representation is not a valid JSON one, but displaying null for those fields is also not valid. Might it be possible to display Infinity/-Infinity/NaN as string instead of null? (the representation will still be wrong, but we may be able to distinguish the different cases)

jeromevalentin:
https://github.com/mongo-express/mongo-express/issues/391

### DIFF
--- a/lib/bson.js
+++ b/lib/bson.js
@@ -120,6 +120,8 @@ exports.toString = function (doc) {
 exports.toJsonString = function (doc) {
   let sJson = json.stringify(doc, null);
   sJson = sJson.replace(/ObjectID\(([^)]+)\)/g, '{ "$oid": $1 }');
+  sJson = sJson.replace(/([-])?Infinity/g, '"$1Infinity"');
+  sJson = sJson.replace(/NaN/g, '"NaN"');
   sJson = sJson.replace(/DBRef\(([^)]+)\)/g, '{ "$ref": $1 }');
   sJson = sJson.replace(/ISODate\(([^)]+)\)/g, '$1');
   return sJson;
--- a/lib/json.js
+++ b/lib/json.js
@@ -124,10 +124,13 @@ function str(key, holder) {
       return quote(value);
 
     case 'number':
-
-      // JSON numbers must be finite. Encode non-finite numbers as null.
-
-      return isFinite(value) ? String(value) : 'null';
+      // Whatever the value number is:
+      // * finite
+      // * NaN
+      // * +Infinity
+      // * -Infinity
+      // let's returning it as string
+      return String(value);
 
     case 'boolean':
     case 'null':


MikeyBurkman:
Hmm... `Infinity` will be displayed as the string `"Infinity"` in the JSON, which is definitely better than displaying null,, but I see an issue if you try to update anything on that record. Mongo Express currently replaces the record with the JSON/BSON in the text area, so you'll replace your Infinity number value with a string.

Of course, the current behavior is really no better -- currently now if you update anything on the record, you're going to replace the value with null. So we improved it, but I feel it's still not quite as ideal as we can make it...

I was first thinking we could maybe handle these as special values, like we do for ObjectId, ISODate, Binary, etc. But these aren't special BSON values... And then that got me thinking: this doesn't have to be valid JSON. Obviously things like ObjectId are not JSON. It needs to be valid BSON. Maybe we can just display +Infinity, etc as-is, NOT as a string? I think Mongo would handle that? @dozoisch ?

dozoisch:
@MikeyBurkman we can definitely handle them as we want. We just need to find what is the format needed for it to be valid for Mongo. It might cause issues with the "export as json". I don't recall on top of my head what we did there for the other invalid values (like ObjectID), but I think that's a good idea.

MikeyBurkman:
@dozoisch Ah yeah, the export is a good point. When we export as JSON, ObjectId gets turned into something like `"_id":{ "$oid": "5aa05dc597adef8018fdc747" }` and ISODate gets turned into its toISOString() equivalent. I'm not totally sure where this conversion happens?

I guess there's a difference between what we display in the editor, and what we export. The exported stuff has to be valid JSON, and it's not re-importable without changing the data quite a lot as a result. What's displayed in the editor has to be valid BSON and should ideally represent exactly what Mongo has stored.

jeromevalentin:
With that PR the stored data in mongo are correct, meaning it does not convert NOT finite number to string as we can expect by reading the code but keep the NaN and Infinity values as it.

I created a document through the mongo-express editor like this:
```javascript
{
    "_id": ObjectID("5aa6da1f4b77511670503cfe"),
    "toto": Infinity,
    "tutu": -Infinity,
    "tata": NaN,
    "string": "This is a string",
    "array": [
        NaN,
        Infinity,
        -Infinity
    ]
}
```

And below is what I have really now in mongo:
```
> db.zz.find()
{ "_id" : ObjectId("5aa6da1f4b77511670503cfe"), "toto" : Infinity, "tutu" : -Infinity, "tata" : NaN, "string" : "This is a string", "array" : [ NaN, Infinity, -Infinity ] }
```
But you are right I still have some issue on export which is producing an invalid JSON document
```javascript
{"_id":{ "$oid": "5aa6da1f4b77511670503cfe" },"toto":Infinity,"tutu":-Infinity,"tata":NaN,"string":"This is a string","array":[NaN,Infinity,-Infinity]}
```
which is not a valid JSON object as NaN, +/-Infinity are not valid values.

Does someone know where the JSON export can be customized? And what is the standard way to export such values in JSON? as String? or as null?

MikeyBurkman:
A ha, I was misunderstanding the point of the String constructor call. It's converting it to a JS string, not a JSON string -- `'Infinity'` vs `'"Infinity"'`. So that's good. And my concern about editing the record and accidentally changing the data types is now unfounded.

As for the JSON export, I'm ok with exporting the values as `"Infinity"`. 

@dozoisch I'm cool with merging this if you are.

dozoisch:
If the edit works, I'm good with it :)

jeromevalentin:
@dozoisch Does the fact that I've created the document directly from the edit is enough to validate it? Or is there another specific test to perform?

MikeyBurkman:
@jeromevalentin I created the record using the edit popup, then went back, re-opened it, edited it a bit more, and everything still looked good. The fact that you could create a new record using `Infinity` meant that BSON has no trouble with handling those values, and that was my biggest concern. 

I suspect when the code was first written, it was trying to stick to pure JSON, not BSON, hence the original null conversion.

MikeyBurkman:
@jeromevalentin Also, nothing to worry about with this PR (unless you want), but it looks like your local git config is not configured with your github email, so your commits aren't linked to your github profile. Note the lack of clickable profile pic next to your commits. As a result, I don't think these contributions will show up in github correctly. 

If you want, you can probably rebase the commits after fixing your github config, and the new rebase commit should then be under your profile. If you don't care one way or another then I'll just merge this one :)

jeromevalentin:
@MikeyBurkman I added the right email address to my profile, does that fix the 'configuration' issue?

MikeyBurkman:
I think maybe the username is also required? In my .gitconfig file I have this:
```
[user]
        name = MikeyBurkman
        email = <my email>
        password = <hashed password>
```

jeromevalentin:
@MikeyBurkman It's ok, integrates it as it :)

dozoisch:
I'll try to cut a release tonight.

PR #396, opened 2018-03-27
Fix bug in filters that mistakenly converts non-Binary and non-Object…

bnielsen1965:
…ID objects into the string '[object Object]'.

The current filters.js is converting all nested objects into the string "[object Object]" in the collection view. This change to the filters.js makes it so any objects that are not Binary nor ObjectID will be returned untouched.

### DIFF
--- a/lib/filters.js
+++ b/lib/filters.js
@@ -148,8 +148,9 @@ exports.stringDocIDs = function (input) {
       case 'Binary':
         return input.toJSON();
       case 'ObjectID':
-      default:
         return input.toString();
+      default:
+        return input;
     }
   }
 


dozoisch:
Hey @bnielsen1965 thanks a lot, I'll try to get to it by tonight, and cut a release

MikeyBurkman:
@bnielsen1965 Can you give an example or screenshot of where you're seeing this, or explain how to reproduce the issue? Nested objects in the collection view seem to display alright to me, so I'm probably missing something.

bnielsen1965:
@MikeyBurkman sure thing, the following are screenshots before and after the fix...

objectbug

objectfix

MikeyBurkman:
@bnielsen1965 Yep, I was missing something alright. I was somehow running an old version of the code that didn't have that regression! Your fix looks good to me.

dozoisch:
@MikeyBurkman @bnielsen1965 tagged and published 0.46.1!

PR #410, opened 2018-05-06
[Fix #409] update convertBytes filters in filters.js

### LINKED ISSUE #409: average doc size shows "NaN Bytes" in empty collection page
average object size display as nan

I think this happened because the stats object does not return the property "avgObjSize"

jovehammer16:
Fix #409 
Might not be the best solution but at least do not display a "NaN" value on the page


dozoisch:
thank you!

PR #419, opened 2018-06-05
Export collections with filenames encoded according to RFC 6266

### LINKED ISSUE #418: Collection names with non ASCII characters cause "Internal Server Error"
to reproduce:
On mongo shell, create a collection names with non ASCII characters.
```
db.テスト.insert({})
```
View the collection with a web browser and click "Export" button.

server side error log:
```
TypeError: The header content contains invalid characters
    at validateHeader (_http_outgoing.js:494:11)
    at ServerResponse.setHeader (_http_outgoing.js:498:3)
    at exp.exportCollection (/app/lib/routes/collection.js:317:9)
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
    at next (/app/node_modules/express/lib/router/route.js:131:13)
    at mongoMiddleware (/app/lib/router.js:282:5)
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
    at next (/app/node_modules/express/lib/router/route.js:131:13)
    at Route.dispatch (/app/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
GET /db/test/export/%E3%83%86%E3%82%B9%E3%83%88 500 4.118 ms - 22
```

hryknkgw:
Fix #418

### DIFF
--- a/lib/routes/collection.js
+++ b/lib/routes/collection.js
@@ -314,8 +314,11 @@ var routes = function (config) {
   };
 
   exp.exportCollection = function (req, res) {
-    res.setHeader('Content-disposition', 'attachment; filename=' + req.collectionName + '.json');
-    res.setHeader('Content-type', 'application/json');
+    res.setHeader(
+      'Content-Disposition',
+      'attachment; filename="' + encodeURI(req.collectionName) + '"; filename*=UTF-8\'\'' + encodeURI(req.collectionName)
+    );
+    res.setHeader('Content-Type', 'application/json');
 
     var query;
     var query_options;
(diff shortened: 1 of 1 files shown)


MikeyBurkman:
@dozoisch This kind of raises another issue -- mongo-express doesn't actually allow you to _create_ a collection with non-English characters. If Mongo allows it, should we change our regex to allow it as well? 

Right now we do `if (!name.match(/^[a-zA-Z_/-][a-zA-Z0-9._/-]*$/))`, but maybe we should just instead look for known illegal characters? The Mongo docs explicitly give `/\. "$*<>:|?`. I looked back to the Mongo 2.4 docs, and the only difference between that and current is it didn't have `$` allowed in the collection name, so I don't think we have to worry much about backwards compatibility.

dozoisch:
@MikeyBurkman that seems like a good approach.

PR #425, opened 2018-07-03
bump a few packages to clear audit

CristiCimpianu:
express in particular seems to draw in a high priority vulnerability


dozoisch:
@CristiCimpianu thanks! I feel like the mocha bump is creating an issue in CI.

PR #456, opened 2019-01-31
Fix importing file bug

mike442144:
If we import file which is exported by mongo-express, we'll get 400 error. Fix the bug by replacing the mimetype.


dozoisch:
@mike442144 hey thank you for following up! I'll get this merged by Saturday

dozoisch:
release incoming shortly, sorry about delay, been swamped with work.

PR #462, opened 2019-04-02
Fix default values of BASICAUTH environment variables in README

dimitarnestorov:


dozoisch:
Hi @dimitarnestorov 
The default values are set here: https://github.com/mongo-express/mongo-express/blob/master/config.default.js#L133-L134 which is what the readme is based on.

dimitarnestorov:
When deploying the Docker container without providing any environment variables there is no HTTP auth required to use mongo-express. This is caused by the `_FILE` variables defined here:
https://github.com/mongo-express/mongo-express/blob/2a8af9f1978c6bce899cfa872fd85f70fbfc776e/Dockerfile#L27-L28

The section with environment variables is listed under the Docker section and having those defaults but no HTTP authentication required created confusion on my end. I was expecting to have to use `admin:pass` to login.

Another suggestion is to move the environment variables section in its own section. And maybe a quick write-up for the Docker section about the `_FILE` variables.

dozoisch:
@dimitarnestorov ah that seems more like a bug in the docker config then

PR #473, opened 2019-07-09
Merge for better UUID

damien-malescot:


dozoisch:
Thank you! Will review this tonight or tomorrow night :)

dozoisch:
@damien-malescot haven't forgot about this, been a bit of a crazy week :) Sorry for delay

damien-malescot:
Hi, 
I didn't see that build was failed (lint).
Just fix it ;)

Another question : is docker image automatically tag after merge ?

dozoisch:
@damien-malescot the image is a manual process

damien-malescot:
Ok need PR to update it ?

dozoisch:
Merge the two other PR, will merge this one tomorrow morning for real :)

damien-malescot:
Hehe no prob, take our time 😉

dozoisch:
@damien-malescot So I noticed that `new Buffer(string, encoding)` is deprecated. 

https://nodejs.org/api/buffer.html#buffer_new_buffer_string_encoding

Can you change the call to `Buffer.from(id, 'base64')` instead.


Also is there a way we can differentiate between base3 and 4 given what is passed? I'm not sure comfortable doing 3 ors in the find it looks a little too much, especially given that normal case will be obj_id. 

If we can differentiate between 3 and 4 then we don't have to look for all 3 of them we take the valid one only. 

Otherwise I guess we could have:

```
const findArgs = {};
if (obj_id) {
  findArgs._id = obj_id;
} else {
 // If we can differentiate here then we don't need to look for both each time
 findArgs["$or"] = [{ _id: binData3 }, { _id: binData4 }];
}

req.collection.findOne(findArgs, ...)
```

damien-malescot:
Ok thx for review.

I will check tomorrow if we can differentiate both or not and pull new commit 😉

damien-malescot:
No solution (for me) to differentiate bindata subtype from base64 value.

So, added a new query param (only for bindata id) when loading document. So, now we have only one argument for _id.

Waiting your review about that solution

lib/router.js
```diff
@@ -223,12 +184,34 @@ const router = function (config) {
...
+    };
+
+    // If an ObjectID was correctly created from passed id param, try getting the ObjID first else falling back to try getting the string id
+    // If not valid ObjectID created, try getting string id
+    req.collection.findOne({ _id: _id }, function (err, doc) {
+      if (err) {
+        req.session.error = 'Error: ' + err;
+        return res.redirect(res.locals.baseHref + 'db/' + req.dbName + '/' + req.collectionName);
+      }
+      console.log(doc);
```
dozoisch: looks like forgotten console.log!

dozoisch:
Thanks for the update, I'll try to review it all tonight =) !

lib/router.js
damien-malescot: Opsss ! Will fix it ;)

dozoisch:
This solution looks much better! Trying to get it merged asap.

dozoisch:
merged 1965f8db092d48851c5f7dea65128ca6dfd84c85

Issue #457, opened 2019-02-22
Is it possible to set the content on full width (front) ?


thib3113:
Actually, the with of the front use the class `container`, is it possible to use `container-fluid` insteed (maybe with a custom marging/padding ) . 

Because, mongo-express will show a table, with some times really lot of columns (where robot 3T and compass prefer using lines), and on some screens, the "usable" interface is really small .

Example with a 21/9 screen : 
![image](https://user-images.githubusercontent.com/5489218/53236947-c551a500-3695-11e9-8e5f-66d1060c3f35.png)

dozoisch:
I think it could be great to have a settings that allow for container-fluid instead of container. 

If you have time to put up a PR that would be really appreciated.

PR #495, opened 2019-08-19
Configurable full width layout using container-fluid vs container #457

Premik:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/495)
---

Although some buttons look odd when they are too wide. Especially if the browser window width is very large like spread across two monitors. Perhaps this could be improved further by adding some `maxWidth` or/and `flex-grow` etc. 
FullWidth


Premik:
I'm closing this one since the #919 has been merged..

PR #504, opened 2019-09-10
Big comparison

slowdown2014:
Able to merge


dozoisch:
@slowdown2014 the gridfs branch is behind master and has no commit that are not in master. I'm not sure what would be the need to  merge this back into gridfs.

PR #517, opened 2019-11-01
Bumped Travis CI to run on Node 6 / 8 / 10 / 12 / 13

mathieumg:
Wasn't sure if you wanted to keep 6 and maybe even 8 (technically still in maintenance for a few months), can update accordingly.


dozoisch:
We can keep it for now I think! Thanks

PR #522, opened 2019-11-11
Change bson parsing to use mongo-query-parser

dozoisch:


patches/mongodb-query-parser+1.4.3.patch
```diff
@@ -0,0 +1,19 @@
+diff --git a/node_modules/mongodb-query-parser/lib/index.js b/node_modules/mongodb-query-parser/lib/index.js
```
JLLeitschuh: Did you intend to leave this file checked in?

dozoisch: @JLLeitschuh yep, have to as query-parser was missing support for the $db parameter, and Timestamp was using the new bson format, while the dependency was still using the old format, so needed to make sure they were consistent

lib/bson.js
```diff
@@ -1,65 +1,11 @@
 'use strict';
 
-var mongodb = require('mongodb');
-var vm      = require('vm');
-var json    = require('./json');
+const parser = require('mongodb-query-parser');
+const ESJON = require('mongodb-extended-json');
```
mathieumg: https://github.com/mongodb-js/mongodb-extjson#note-this-libray-has-been-merged-into-the-4x-version-of-the-bson-package-and-is-no-longer-being-maintained-as-a-standalone-package

https://github.com/mongodb/js-bson#ejsonparsetext-options

dozoisch: Unfortunately, can't update bson until mongodb-query-parser is updated to a more recent version.

Issue #512, opened 2019-10-20
Import fails: "Bad file"


kuba-orlik:
Importing a collection dump exported from another instance of mongo-express has no result. No error show up, but in the Network Inspector I can see there's a `400` error with the only response content being 

```
Bad file
```

moomoo-dev:
Was this ever fixed ?

rtritto:
> Was this ever fixed ?

@moomoo-dev yes (source: https://github.com/mongo-express/mongo-express/issues/1225#issuecomment-1701077372)

Issue #433, opened 2018-08-30
import does not work on json arrays files.


positron96:
Gives 
```
{ MongoError: operation passed in cannot be an Array
mongo-express_1  |     at Function.MongoError.create (/node_modules/mongodb-core/lib/error.js:31:11)
mongo-express_1  |     at toError (/node_modules/mongodb/lib/utils.js:114:22)
mongo-express_1  |     at addToOperationsList (/node_modules/mongodb/lib/bulk/unordered.js:182:11)
mongo-express_1  |     at UnorderedBulkOperation.raw (/node_modules/mongodb/lib/bulk/unordered.js:387:7)
mongo-express_1  |     at bulkWrite (/node_modules/mongodb/lib/collection.js:636:12)
mongo-express_1  |     at /node_modules/mongodb/lib/collection.js:530:5
mongo-express_1  |     at new Promise (<anonymous>)
mongo-express_1  |     at Collection.insertMany (/node_modules/mongodb/lib/collection.js:529:10)
mongo-express_1  |     at Collection.insert (/node_modules/mongodb/lib/collection.js:825:15)
mongo-express_1  |     at keys.forEach (/node_modules/mongo-express/lib/routes/collection.js:582:24)
mongo-express_1  |   name: 'MongoError',
mongo-express_1  |   message: 'operation passed in cannot be an Array',
mongo-express_1  |   driver: true }
```

I have latest docker image of mongo-express (0.49).

Actually, there are 3 export buttons in the UI and it not clear what should I pass into import. Anyway, I tried all of them, none work. json array gives most info on console, others silently give http 400.

wakabayashik:
https://github.com/mongo-express/mongo-express/blob/797e68de25e3566f658b3691ddbb2aac20e393c0/lib/routes/collection.js#L576-L577

I modified collection.js in my working clone and it works for me...
```javascript
var fileContent = req.files[key].data.toString('utf8');
if (fileContent.slice(0,1) === '[' && fileContent.slice(-1) === ']') fileContent = fileContent.slice(1,-1);
const str = `[${fileContent.replace(/}[\r\n]{1,2}(\s)?\{/gm, '},{').replace(/\{\s*"\$oid"\s*:\s*"(.+?)"\s*}/gm, '"$1"')}]`;
```

smn-snkl:
Is this working now? I still can't get imports to work ...

PabloSzx:
[...]

PR #534, opened 2019-11-18
fix import and export

kambing86:
fix import and export issues
https://github.com/mongo-express/mongo-express/issues/512
https://github.com/mongo-express/mongo-express/issues/433


kambing86:
closed because `bson` has issue when stringify `Binary`

https://github.com/mongodb/js-bson/issues/331

dozoisch:
Thanks for the follow-up, I subscribed to 331 to see when this gets resolved.

Issue #550, opened 2020-01-21
Open browser on start


aeberdinelli:
It would be nice to open the browser in mongo-express when the app has started.

PR #551, opened 2020-01-21
Open in browser

aeberdinelli:
As described on issue #550 

Open browser on startup, uses the [open](https://npmjs.com/package/open) package and can be enabled using the config file.


MikeyBurkman:
I'm not opposed to this feature per se, but it definitely shouldn't be on by default, given that this app is typically (or at least often) run on a server, not locally. We also probably don't want to change the current default behavior.

Will defer to other maintainers whether they think this is a worthwhile feature to add or not.

aeberdinelli:
@MikeyBurkman I thought about that as well, I think I should switch it to off by default, I will update the PR

PR #575, opened 2020-05-14
GPG key check in dockerfile broken (?), removed

jotelha:
Was not able to build Docker image without removing the key key check. Keyserver ha.pool.sks-keyservers.net yields No results found: No keys found.

---
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/575)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,10 +7,6 @@ RUN set -x \
 		--no-install-recommends \
 	&& curl -fSL "https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini" -o /usr/local/bin/tini \
 	&& curl -fSL "https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini.asc" -o /usr/local/bin/tini.asc \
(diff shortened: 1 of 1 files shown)


bwitt:
I think it's still a good idea to validate the downloaded files; I notice the newer versions e.g. https://github.com/krallin/tini/releases/tag/v0.19.0 have sha256sums we could validate instead of using gpg

bwitt:
oh actually, we could just verify the current version file now via sha256sum with a hardcoded sha256 instead of gpg e.g.
```
&& echo '<output of sha256sum> /usr/local/bin/tini' | sha256sum --check 
```

dungmv:
consider changing keyserver
https://stackoverflow.com/questions/66217436/gpg-keyserver-receive-failed-no-name

BlackthornYugen:
@jotelha #804 should have corrected this. Can you verify if the updated dockerfile works for you?

BlackthornYugen:
I'll close this for now. If there's still an issue we can use #842 to track it.

Issue #581, opened 2020-07-03
"UnhandledPromiseRejectionWarning: MongoNetworkError: failed to connect to server [undefined:27017] on first connect [Error: getaddrinfo ENOTFOUND undefined"


tianon:
Hi!  When trying to update https://github.com/mongo-express/mongo-express-docker to 1.0.0-alpha.1, I can't seem to run `mongo-express` successfully no matter how I install version 1.0.0-alpha.1, and I hope it's because there's something obvious I'm doing wrong.

No matter what I do (removing all environment variables, installing via `yarn` vs `npm`, etc etc etc), I'm getting the following error/backtrace:

```console
No custom config.js found, loading config.default.js
Welcome to mongo-express
------------------------


(node:1) DeprecationWarning: current Server Discovery and Monitoring engine is deprecated, and will be removed in a future version. To use the new Server Discover and Monitoring engine, pass option { useUnifiedTopology: true } to the MongoClient constructor.
Could not connect to database at index "0"
(node:1) UnhandledPromiseRejectionWarning: MongoNetworkError: failed to connect to server [undefined:27017] on first connect [Error: getaddrinfo ENOTFOUND undefined
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:66:26) {
  name: 'MongoNetworkError'
}]
    at Pool.<anonymous> (/opt/mongo-express/node_modules/mongodb/lib/core/topologies/server.js:438:11)
    at Pool.emit (events.js:315:20)
    at /opt/mongo-express/node_modules/mongodb/lib/core/connection/pool.js:562:14
    at /opt/mongo-express/node_modules/mongodb/lib/core/connection/pool.js:995:11
    at /opt/mongo-express/node_modules/mongodb/lib/core/connection/connect.js:31:7
    at callback (/opt/mongo-express/node_modules/mongodb/lib/core/connection/connect.js:264:5)
    at Socket.<anonymous> (/opt/mongo-express/node_modules/mongodb/lib/core/connection/connect.js:294:7)
    at Object.onceWrapper (events.js:422:26)
    at Socket.emit (events.js:315:20)
    at emitErrorNT (internal/streams/destroy.js:92:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:60:3)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)
[...]

PR #583, opened 2020-07-03
Update/simplify Dockerfile

### LINKED ISSUE #842: gpg: keyserver receive failed: Server indicated a failure
Hi all,

I pulled the latest master branch and try to build the docker image. But While building I am getting the following error 

image
Kindly let me know what exactly we need to change in docker file to make the docker working as expected

tianon:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/583)
---

This updates the Dockerfile to explicitly build from the "Debian Buster" slim variant of the Node.js official image, and trims down the set of overridden environment variables.

This also updates from "npm" to "yarn" and changes to installing "tini" from Debian instead of downloading it by hand.

I can't currently test this successfully due to https://github.com/mongo-express/mongo-express/issues/581, but I have verified that it builds successfully (which is more than can be said for what currently exists in master, since `node:12-slim` updated to be based on `debian:buster` instead of `debian:stretch` and thus no longer has `gpg` installed).

If we can figure out what's causing https://github.com/mongo-express/mongo-express/issues/581 and this needs changes, I'll happily apply them. :smile:

As always, happy to rebase/amend/adjust/close as desired. :+1:

---

Fix #842

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,5 @@
 FROM node:18-alpine3.16
 
-# grab tini for signal processing and zombie killing
-RUN apk -U add --no-cache bash tini
-
-EXPOSE 8081
-
 # "localhost" doesn't mean much in a container, so we adjust our default to the common service name "mongo" instead
 # (and make sure the server listens outside the container, since "localhost" inside the container is usually difficult to access)
 ENV ME_CONFIG_MONGODB_URL="mongodb://mongo:27017" \
@@ -13,7 +8,13 @@ ENV ME_CONFIG_MONGODB_URL="mongodb://mongo:27017" \
 
 WORKDIR /opt/mongo-express
 COPY . .
-RUN yarn install
-# RUN yarn run build	# prepublish already run build
 
+RUN apk -U add --no-cache \
+        bash \
+        # grab tini for signal processing and zombie killing
+        tini \
+    && yarn install
+    # && yarn run build     # prepublish already run build
+
+EXPOSE 8081
 CMD ["/sbin/tini", "--", "yarn", "start"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,52 +1,21 @@
 FROM node:18.7.0-slim
 
 # grab tini for signal processing and zombie killing
-ENV TINI_VERSION 0.9.0
-RUN set -x \
-	&& apt-get update && apt-get install -y ca-certificates curl \
-		--no-install-recommends \
-	&& apt-get install -y gpg \
-	&& curl -fSL "https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini" -o /usr/local/bin/tini \
-	&& curl -fSL "https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini.asc" -o /usr/local/bin/tini.asc \
-	&& export GNUPGHOME="$(mktemp -d)" \
-	&& key=6380DC428747F6C393FEACA59A84159D7001A4E5 \
-	&& ( gpg --batch --keyserver hkps://keyserver.ubuntu.com --recv-keys "$key" || gpg --batch --keyserver hkps://keys.openpgp.org --recv-keys "$key" ) \
-	&& gpg --batch --verify /usr/local/bin/tini.asc /usr/local/bin/tini \
-	&& rm -r "$GNUPGHOME" /usr/local/bin/tini.asc \
-	&& chmod +x /usr/local/bin/tini \
-	&& tini -h \
-	&& apt-get purge --auto-remove -y ca-certificates curl \
-	&& rm -rf /var/lib/apt/lists/*
+RUN set -eux; \
+	apt-get update; \
+	apt-get install -y --no-install-recommends tini; \
+	rm -rf /var/lib/apt/lists/*
 
 EXPOSE 8081
 
-# override some config defaults with values that will work better for docker
-ENV ME_CONFIG_EDITORTHEME="default" \
-    ME_CONFIG_MONGODB_URL="mongodb://mongo:27017" \
-    ME_CONFIG_MONGODB_ENABLE_ADMIN="true" \
-    ME_CONFIG_BASICAUTH_USERNAME="" \
-    ME_CONFIG_BASICAUTH_PASSWORD="" \
-    ME_CONFIG_BASICAUTH_USERNAME_FILE="" \
-    ME_CONFIG_BASICAUTH_PASSWORD_FILE="" \
-    ME_CONFIG_MONGODB_ADMINUSERNAME_FILE="" \
-    ME_CONFIG_MONGODB_ADMINPASSWORD_FILE="" \
-    ME_CONFIG_MONGODB_AUTH_USERNAME_FILE="" \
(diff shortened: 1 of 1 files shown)


dozoisch:
Thanks a lot @tianon I'll try to take a look over the weekend. Very swamped with work right now :)

Dockerfile
```diff
@@ -1,50 +1,20 @@
...
     ME_CONFIG_MONGODB_ENABLE_ADMIN="true" \
-    ME_CONFIG_BASICAUTH_USERNAME="" \
-    ME_CONFIG_BASICAUTH_PASSWORD="" \
-    ME_CONFIG_BASICAUTH_USERNAME_FILE="" \
-    ME_CONFIG_BASICAUTH_PASSWORD_FILE="" \
-    ME_CONFIG_MONGODB_ADMINUSERNAME_FILE="" \
-    ME_CONFIG_MONGODB_ADMINPASSWORD_FILE="" \
-    ME_CONFIG_MONGODB_AUTH_USERNAME_FILE="" \
-    ME_CONFIG_MONGODB_AUTH_PASSWORD_FILE="" \
-    ME_CONFIG_MONGODB_CA_FILE="" \
```
tianon, line 33: Ahhh, the table of config values in the README is reflecting this list, not the defaults that are pre-baked into `config.default.js` (for example, `ME_CONFIG_BASICAUTH_USERNAME` defaults to `admin`, not `""` :sweat_smile:).

Do you think it would make sense to adjust some of these defaults to match `config.default.js` (and update the README as well), or do you think it would make more sense to update `config.default.js` to match what the documented defaults are?

(Happy to adjust this either/neither way. :+1:)

knickers: I think I'd like to convert the docker image to follow the defaults in the config, I've kinda regretted giving the docker image different default values :grimacing:

dozoisch:
@knickers hey!  I see that you reviewed the PR, what's your opinion on it? If you say it's good to go, I'm down to merge it.

knickers:
Hey @dozoisch 😊 I think this PR looks good, but we haven't been able to test because of a startup error, documented in #581. We could use some help deciphering the error if you have a minute 👀

tianon:
It worked (the fix in #614)! :smile:

If I set `ME_CONFIG_MONGODB_URL=mongodb://mongo` as a default in the `Dockerfile`, it works, and `ME_CONFIG_MONGODB_SERVER` then takes precedence if users are specifying that older variable, but now I'm seeing why we'd set so many separate defaults for Docker -- it's complaining about database authentication. :sweat_smile:

From what I can tell reading the following bit of code, the username/password seem to be defaulting to literally `ME_CONFIG_MONGODB_AUTH_USERNAME` if the variables aren't set at all, which I don't think is quite right, and is fixed with the minor adjustment I've now included here.

Also, it only shows one database by default, which is why we'd likely defaulted `ME_CONFIG_MONGODB_ENABLE_ADMIN` to `true` previously. :innocent:
(Although after adding that, I get all my databases, but the dashboard still says "Turn on admin in config.js to view server stats!" which is interesting -- maybe it's also expecting `ME_CONFIG_MONGODB_ADMINUSERNAME` or something?  It feels like a missing assumption somewhere, but definitely not critical.)

So with my latest push, this seems to be working (even for users specifying still `ME_CONFIG_MONGODB_SERVER` or wanting to use `ME_CONFIG_MONGODB_URL` with minimal friction), but there is the open question of whether we should consider to override `ME_CONFIG_MONGODB_ENABLE_ADMIN=true` for Docker. :smile:

tianon:
I guess the case this *doesn't* solve is users specifying `ME_CONFIG_MONGODB_ADMINUSERNAME` and relying on the default server name of `mongo`, but I'd imagine that's a minority.

tianon:
Thanks to https://github.com/mongo-express/mongo-express/pull/710 this works now! :smile:

shakaran:
@tianon could you rebase and resolve the conflicts for see a soon merge? Thanks

shakaran:
Closing this as fixed by #710

rtritto:
Changes of this PR:
-  removes this env vars:
```config
ME_CONFIG_MONGODB_ENABLE_ADMIN="true"
ME_CONFIG_BASICAUTH_USERNAME=""
ME_CONFIG_BASICAUTH_PASSWORD=""
ME_CONFIG_BASICAUTH_USERNAME_FILE=""
ME_CONFIG_BASICAUTH_PASSWORD_FILE=""
ME_CONFIG_MONGODB_ADMINUSERNAME_FILE=""
ME_CONFIG_MONGODB_ADMINPASSWORD_FILE=""
ME_CONFIG_MONGODB_AUTH_USERNAME_FILE=""
ME_CONFIG_MONGODB_AUTH_PASSWORD_FILE=""
ME_CONFIG_MONGODB_CA_FILE=""
```
Maybe the only usefull env var to leave is _ME_CONFIG_MONGODB_ENABLE_ADMIN_ (default is _false_) or is better to change default to _true_
- Replace Node image from _slim_ (~75 MB) to _buster-slim_ (~70 MB):
  - in _mongo-express-docker_ repo was used _alpine_ (~50 MB) ---> _alpine_ should be enough
- Simplify RUN _codes_

@BlackthornYugen @shakaran we should reconsider...

rtritto:
Rebased.
@BlackthornYugen @shakaran did some changes, but container is locked:
```console
$ cross-env NODE_ENV=production node app.js
No custom config.js found, loading config.default.js
Welcome to mongo-express
------------------------
```

Note: to test Dockerfile, we should use this instead of the one of _mongo-express-docker_

shakaran:
> 

See this change in my fork for the config.js file

https://github.com/shakaran/mongo-express-docker/blob/master/Dockerfile#L30

It is related to WORKDIR

rtritto:
> > 
> 
> See this change in my fork for the config.js file
> 
> https://github.com/shakaran/mongo-express-docker/blob/master/Dockerfile#L30
> 
> It is related to WORKDIR

I try to run _start_ script on my machine and I have same result.

rtritto:
@shakaran 
I missed some env vars, now it works!
With _COPY . ._, all files are copied to current directory, so _RUN cp ..._ for _config.js_ file should not be needed.

For me, this can be merged.
FYI @BlackthornYugen

MahdiAbbasi95:
> @shakaran I missed some env vars, now it works! With `COPY . .`, all files are copied to current working directory, so `RUN cp ... ...` command for _config.js_ file should not be needed.
> 
> For me, this can be merged. FYI @BlackthornYugen

I think it's better we merge two RUN in the Dockerfile. It's best practice and it will decrease one layer.
Please check my fork:

https://github.com/MahdiAbbasi95/mongo-express/blob/dockerfile/Dockerfile

rtritto:
@MahdiAbbasi95 thanks, I did a commit.

rtritto:
Fix #842

BlackthornYugen (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/583)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tianon)

PR #613, opened 2020-10-01
Added support for running raw commands on a collection

slasktrat:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/613)
---

Thanks for this great tool! A feature I've been missing for a while is to be able to run raw commands on a collection. To e.g. updateMany etc. I didn't dive deep into the sctructure of the app, so I might not have implemented it ideally but it works :)

This PR creates a new Tab in the collection view where the user can run raw commands. Syntax in the input field would be like: updateMany({myproperty:"someValue"}, {$set:{myproperty:"someOtherValue"}})

The tab is hidden if either ME_CONFIG_OPTIONS_NO_RAW_COMMAND or ME_CONFIG_OPTIONS_READONLY is set to true

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -62,6 +62,7 @@ module.exports = {
     "prefer-const": ["off"],
     "object-shorthand": ["off"],
     "no-path-concat": ["off"],
+    "no-eval": ["off"],
     "no-shadow": ["off", {allow: ["err", "error"]}],
     "camelcase": ["off"],
   },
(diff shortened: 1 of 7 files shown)


armellarcier:
How comes there is no discussion nor issues linked ?

shakaran:
@slasktrat could you rebase your PR and resolve conflicts for a easy merge from the maintainer? Thanks

BlackthornYugen:
@rtritto @slasktrat @shakaran maybe a better idea is to support "plugins" for mongo-express, and we can have a plugin that allows raw commands that some users can choose to enable -- but I worry that it's too risky to officially support code injection.

rtritto:
> @rtritto @slasktrat @shakaran maybe a better idea is to support "plugins" for mongo-express, and we can have a plugin that allows raw commands that some users can choose to enable -- but I worry that it's too risky to officially support code injection.

This feature inject codes into app (I think it's unsafe). In UI maybe we can limit actions/options by some dropdown.

Closed in favor of #939

PR #614, opened 2020-10-04
+ Fix invalid default values for config

JafarAkhondali:
I think this will fix the undefined values for the host, however, the config file needs a proper refactoring
fix: #581

You may also add `hacktoberfest` topic in your project to attract other contributors


dozoisch:
Thanks @JafarAkhondali added it to the topics as well :).

PR #615, opened 2020-10-04
Fix query options bug from Mongodb breaking change

JafarAkhondali:
The PR: #558 updated mongodb version from `2.2.24` to `3.5.5`. There are some breaking changes in mongodb 3.x.x version like [find](https://github.com/mongodb/node-mongodb-native/blob/HEAD/CHANGES_3.0.0.md#find) which removed the projection field as separate parameters and expectes it in options.  
Since projection was the second field and query_options was the third one, all features like sorting and pagination were corrupted by the update. This PR will fix this issue.

### DIFF
--- a/lib/routes/collection.js
+++ b/lib/routes/collection.js
@@ -93,16 +93,12 @@ var routes = function (config) {
   };
 
   exp._getQueryOptions = function (req) {
-    var limit = config.options.documentsPerPage;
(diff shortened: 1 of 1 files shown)


dozoisch (approved):

dozoisch:
Thanks!

PR #704, opened 2021-05-23
Fix document delete on collection view for UUID (BinData 3 & 4)

### LINKED ISSUE #668: Unable to delete doc with UUID as _id
Document not found appears when I try to delete a doc that uses UUID instead of ObjectId.

I use UUIDs because it looks like a random ID, so the user cannot guess easily the next id.

denebgarza:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/704)
---

This will fix #668 

Request path is missing `subtype` param. Figured this after looking at this commit for reference https://github.com/mongo-express/mongo-express/pull/538/commits/760a14f5adc559577fa5e568b7a01aa1be8ea0e8

Turns out this issue had been fixed previously, but only in the individual document view and not the collection view.

### DIFF
--- a/lib/views/collection.html
+++ b/lib/views/collection.html
@@ -263,7 +263,7 @@
           {% for column in columns %}
             <td><div class="tableContent">
               {% if !settings.read_only && !settings.no_delete && column === '_id' && collectionName !== 'system.indexes' %}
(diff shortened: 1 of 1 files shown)


BlackthornYugen (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/704)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @denebgarza)

PR #738, opened 2021-07-21
Fix default config.

sky0014:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/738)
---

When `ME_CONFIG_MONGODB_URL` is set, other envs will be ignored like `ME_CONFIG_MONGODB_PORT`, `ME_CONFIG_MONGODB_AUTH_DATABASE` etc.

So `ME_CONFIG_MONGODB_URL` should not set by default, is better to set by user.

Btw, mongo offical docker hub page used `mongo-express`, and it's example can't work now, this pr should fix this.

from [https://hub.docker.com/_/mongo](https://hub.docker.com/_/mongo)
```yml
# Use root/example as user/password credentials
version: '3.1'

services:

  mongo:
    image: mongo
    restart: always
    environment:
      MONGO_INITDB_ROOT_USERNAME: root
      MONGO_INITDB_ROOT_PASSWORD: example

  mongo-express:
    image: mongo-express
    restart: always
    ports:
      - 8081:8081
    environment:
      ME_CONFIG_MONGODB_ADMINUSERNAME: root
      ME_CONFIG_MONGODB_ADMINPASSWORD: example
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ EXPOSE 8081
 
 # override some config defaults with values that will work better for docker
 ENV ME_CONFIG_EDITORTHEME="default" \
-    ME_CONFIG_MONGODB_URL="mongodb://mongo:27017" \
+    ME_CONFIG_MONGODB_SERVER="mongo" \
     ME_CONFIG_MONGODB_ENABLE_ADMIN="true" \
     ME_CONFIG_BASICAUTH_USERNAME="" \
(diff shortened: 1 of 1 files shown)


OOCAZ:
Potentially could also set up this fix in a .env file like what I suggested with [#983](https://github.com/mongo-express/mongo-express/issues/983)

rtritto:
As said, after #957, this change isn't needed.

FYI @shakaran @BlackthornYugen

PR #815, opened 2021-11-25
CA should be a path to a PEM file

BlackthornYugen:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/815)
---

Fixes #814

PR #851, opened 2022-05-17
Support Node 18

BlackthornYugen:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/851)
---

## ✅ Node 18

It looks like node-18 works without any changes

## ⚠️ Updated Dependency
~~`npm audit fix --force` to update `express-fileupload`, before I take off "draft" I will make sure that file upload still works.~~

**edit**  I've tested that export/import still works with the change

## ~~⚠️ Failing Tests~~

~~28/58 tests fail at the moment on my machine. I am having trouble getting any of the router tests to run:~~

image

**edit**: it seems to work on the CI build

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "csurf": "1.10.0",
     "errorhandler": "1.5.1",
     "express": "4.17.1",
-    "express-fileupload": "^0.4.0",
+    "express-fileupload": "^1.3.1",
     "express-session": "^1.15.6",
     "flat": "^2.0.1",
     "gridfs-stream": "^1.1.1",
--- a/.github/workflows/standard-ci.yml
+++ b/.github/workflows/standard-ci.yml
@@ -15,7 +15,7 @@ jobs:
   test:
     strategy:
       matrix:
-        node: [12, 14, 16]
+        node: [12, 14, 16, 18]
         os: [ubuntu-latest]
     name: Node v${{ matrix.node }} - ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
@@ -27,9 +27,9 @@ jobs:
         - 27017:27017
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
-    - uses: actions/setup-node@v2
+    - uses: actions/setup-node@v3
       with:
         node-version: ${{ matrix.node }}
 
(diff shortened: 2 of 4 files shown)


dozoisch (commented):
> Reviewed 7 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/851)* status: 4 of 9 files reviewed, 1 unresolved discussion (waiting on @BlackthornYugen and @dozoisch)

___
*[`lib/scripts/database.js` line 67 at r1](https://reviewable.io/reviews/mongo-express/mongo-express/851#-N2KGSAYCC_hjOA6Kd6I:-N2KGSAYCC_hjOA6Kd6J:b97b6zn) ([raw file](https://github.com/mongo-express/mongo-express/blob/f088747d45d2b28f158dabb0e910343b6c684432/lib/scripts/database.js#L67)):*
> ```JavaScript
>     })
>       .done(function (res) {
>         console.log(res);
> ```

minor: just making sure that those changes are done on purpose because that is a big change of behavior... In all honesty, I dont remember what shows up in the alert but wanted to point it out

BlackthornYugen:
> making sure that those changes are done on purpose because that is a big change of behavior...

Good point, maybe for now I can just suppress the lint error? or it's just a lint warning and it isn't failing the build, I could also just revert those changes too.

BlackthornYugen:
@dozoisch it might be nicer if the lint changes could be moved into their own PR so I created #852. I guess I broke the build with the PEM change in #815.

package.json
```diff
@@ -38,7 +38,7 @@
     "csurf": "1.10.0",
     "errorhandler": "1.5.1",
     "express": "4.17.1",
-    "express-fileupload": "^0.4.0",
+    "express-fileupload": "^1.3.1",
```
BlackthornYugen, line 41: it probably makes sense to have this in it's own PR too. I like how the other version bumps are in a commit named "bump x from y to z" with a "dependencies" tag so it is easy to spot these in the git history and to update history.md. 

Is #592 the only dependabot PR for fileupload? It seems to be updating more than just fileupload.

rtritto:
Node v12 is no longer maintained.
It must be removed because some new dependencies add support to Node v18 and remove to Node v12 in same release.

BlackthornYugen:
Replaced by #911 #916 and #914

PR #921, opened 2022-09-01
chore: update Express.js to v5

rtritto:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/921)
- - -

[Migrate to v5](https://expressjs.com/en/guide/migrating-5.html)

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -76,7 +76,7 @@ const router = async function (config) {
   };
 
   // View helper, sets local variables used in templates
-  appRouter.all('*', async function (req, res, next) {
+  appRouter.all('(.*)', async function (req, res, next) {
     res.locals.baseHref       = buildBaseHref(req.originalUrl, req.url);
     res.locals.databases      = mongo.getDatabases();
     res.locals.collections    = mongo.collections;


rtritto:
Correctly rebased with _master_.
FYI @shakaran

shakaran:
> Correctly rebased with _master_. FYI @shakaran

it seems that still fail for the pipelines

rtritto:
> > Correctly rebased with _master_. FYI @shakaran
> 
> it seems that still fail for the pipelines

I know, that fix works only with express v5.
My original idea is to wait a stable/rc express v5 to update the dependency. By this way, we should have less conflicts with rebases/synch.

shakaran:
@rtritto this need rebase or update maybe

shakaran:
@rtritto any update with this?

rtritto:
v5 isn't updated with v4. Before merge, I would wait for the v5 to become stable.
`express` seems not mantained so we should consider to migrate to `fastify`. I did this branch https://github.com/mongo-express/mongo-express/tree/fastify-migration as draft.

shakaran:
> v5 isn't updated with v4. Before merge, I would wait for the v5 to become stable. `express` seems not mantained so we should consider to migrate to `fastify`. I did this branch https://github.com/mongo-express/mongo-express/tree/fastify-migration as draft.

In that versión, I see that we lost the support for swigEngine, could you add it too and make a PR rebased?

rtritto:
Seems that this changes isn't need anymore with current v5.

Replaced with #1630

PR #938, opened 2022-09-09
Merge Slasktrat/master

shakaran:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/938)
- - -

This is a new branch for issues with merge of #613

PR #939, opened 2022-09-09
feat: add support for running raw commands on a collection

rtritto:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/939)
- - -

Replace #613 (contains some info) and #938


lib/routes/collection.js
```diff
@@ -130,6 +130,43 @@
...
+      try {
+        const match = /^(.*?)\s*\((.*)\)\s*$/.exec(req.query.command);
+        if (!match || !match.length >= 3) {
+          req.session.error = 'Invalid command. Example: updateOne({myproperty:"someValue"}, {$set:{myproperty:"someOtherValue"}})';
+          return res.redirect('back');
+        }
+
+        const fun = match[1];
+        if (typeof req.collection[fun] === 'function') {
+          const args = match[2].split(',').map((s) => eval(`(${s})`));
```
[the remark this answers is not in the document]
rtritto: Current output of ESLint:
```console
143:55  error  eval can be harmful  no-eva
```

I did a try replacing `eval(`(${s})`` with:
```js
new Function(`return (${s})`)()
```
I get:
```console
143:55  error  The Function constructor is eval  no-new-func
```

rtritto:
Current output of ESLint:
```console
143:55  error  eval can be harmful  no-eva
```

I did a try replacing `eval(`(${s})`` with:
```js
new Function(`return (${s})`)()
```
I get:
```console
143:55  error  The Function constructor is eval  no-new-func
```

BlackthornYugen:
This seems pretty dangerous. This should be off by default and we can't trust the UI to disable it.

README.md
```diff
@@ -138,6 +138,7 @@ You can use the following [environment variables](https://docs.docker.com/refere
     `ME_CONFIG_OPTIONS_FULLWIDTH_LAYOUT`    | `false`   | if set to true an alternative page layout is used utilizing full window width.
     `ME_CONFIG_OPTIONS_PERSIST_EDIT_MODE`   | `false`   | if set to true, remain on same page after clicked on Save button
     `ME_CONFIG_OPTIONS_NO_DELETE`      | `false`        | if noDelete is true, components of deleting are not visible.
+    `ME_CONFIG_OPTIONS_NO_RAW_COMMAND`| `false`         | if noRawCommand is true, the Raw tab in collection view is not visible.
```
BlackthornYugen, line 141: Maybe this should be called something else to highlight how risky this is. ME_CONFIG_OPTIONS_ALLOW_CODE_INJECTION? Default must be not to allow it.

shakaran: I agree

shakaran:
Instead of allow this by default, we should only allow with a flag that should be explicitly enabled by the user. So in case of want use "insecure" eval raw commands it is because they choose that option, but this will not be ship as enabled by default

shakaran (approved):
Thanks @rtritto — the feature is a reasonable thing for an admin tool, but the implementation could not ship, so this is a rewrite of the handler with your config option and UI intent kept.

**The blocker was remote code execution.** Arguments were built with

```js
match[2].split(',').map((s) => eval(`(${s})`))
```

so the `command` query parameter ran as Node code on the server. Confirmed with a harmless payload: `find({a: process.version})` evaluated and returned `{"a":"v22.21.1"}`. CodeQL flagged this as code injection on the original commit.

**It also would not have worked on the current stack:**

| | |
|---|---|
| `req.collection[fun](...args, callback)` | the driver has ignored callbacks since v4 and returns a Promise, so the callback never fired and the request hung |
| `res.redirect('back')` | Express 5 sends a literal `Location: back` |
| `!match \|\| !match.length >= 3` | precedence — `(!match.length) >= 3` is always false, so the guard never fired |
| `split(',')` | breaks any argument containing a comma, which most updates do |

**What it does now:**

- The method name is checked against an explicit allow-list, split into read, write and delete groups; anything else, including non-methods, is refused.
- Arguments go through `bson.toSafeBSON`, the parser the query box already uses, with the list wrapped in brackets so it parses as one array — nested documents and commas survive.
- **`readOnly` and `noDelete` are enforced in the handler.** This matters more than it looks: the Raw form submits over GET, so the middleware added in #1900 never sees it, and a raw `deleteMany()` would otherwise walk straight past a read-only instance.
- The result is awaited and cursors drained; errors are reported rather than swallowed.
- Tab markup moved from Bootstrap 3 to 5 to match the rest of the view.

CodeQL's second finding, polynomial backtracking on the command regex, was real. Measured on 60k characters of adversarial input:

| | `' '*60000 + 'a('` | `' '*60000` |
|---|---|---|
| original `/^(.*?)\s*\((.*)\)\s*$/` | **3187 ms** | **2983 ms** |
| replacement | 0.2 ms | 0.1 ms |

**11 tests added.** Restoring the `eval` fails the 'does not evaluate' and comma-argument cases; removing the flag checks fails the `readOnly` and `noDelete` cases.

Both bot comments are anchored to `8040da2`; the head is `530ccd7`. Lint clean, 111 passing, CI green on Node 22, 24 and 26.

PR #995, opened 2022-09-24
added cypress and initial test, still working on code coverage and number of tests, but this is a good initial commit with most every configured

### LINKED ISSUE #967: Need more e2e tests
@rtritto there are unwritten tests in https://github.com/mongo-express/mongo-express/blob/master/test/lib/routers/collectionSpec.js

Do you think we can write tests in this "bootstrap" test file ? What is its purpose ? It seems like we could add test data in `testMongoUtils` and write real tests for different queries

_Originally posted by @armellarcier in https://github.com/mongo-express/mongo-express/issues/951#issuecomment-1247286824_

OOCAZ:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/995)
- - -

Framework for E2E implementation. Will need help making sure the yarn is updated correctly, added several packages.

Fix #967


OOCAZ:
@BlackthornYugen @rtritto Yarn file probably needs to be updated...

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "livePreview.defaultPreviewPath": "c:\\Users\\zacpo\\Documents\\GitHub\\mongo-express\\coverage\\lcov-report\\index.html"
```
shakaran, line 2: Why is this file added?

cypress.config.cjs
```diff
@@ -0,0 +1,18 @@
+const { defineConfig } = require('cypress')
```
shakaran, line 1: We are using ESM with ES6, you need to change the require by import

cypress/e2e/homepage.cy.js
```diff
@@ -0,0 +1,5 @@
+describe('empty spec', () => {
+  it('passes', () => {
+    cy.visit('localhost:8081');
```
shakaran, line 3: Following the .env or NODE_ENV it should load the running local server start, but avoid harcoded values

cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
+// This example commands.js shows you how to
```
shakaran, line 2: Which give us the example commands? We only should keep the used

.vscode/settings.json
OOCAZ: I don't know, Can be delted lol

cypress.config.cjs
OOCAZ: do you know where I need to import require from?

cypress/e2e/homepage.cy.js
OOCAZ: Cypress has it's own variable configs, I will work on that and push a fix

cypress/support/commands.js
OOCAZ: Understood, Cypress created this and I didn't remove, may be fine to remove

cypress.config.cjs
OOCAZ: figured it out

cypress/support/commands.js
OOCAZ: so this file is required for the E2E testing in the event that a custom command is needed

cypress.config.cjs
rtritto: @OOCAZ You must change also the _export_.
It's better to change extension in `.mjs` for ESM ([souce](https://docs.cypress.io/guides/references/configuration#Configuration-File)).

shakaran:
@OOCAZ let me know when you have an update on this and you need rebase with yarn.lock in master

OOCAZ:
> @OOCAZ let me know when you have an update on this and you need rebase with yarn.lock in master

Hoping to get some work done on this later this week. Will tag you here when I need a review.

rtritto:
I can do rebase and the fix `cypress.config` extension file, then I'll test.

rtritto:
Fixed `cypress.config` file, added cypress script and update `.gitignore`.

shakaran (approved):

PR #574, opened 2020-05-14
feat: add two tsl/ssl certificate-related options and adapted lib/db.js

jotelha:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/574)
---

This is rather a note or question than a true pull request. I don't understand much about encryption and nodejs, but I could not get the current mongo-express docker container (0.54) to connect to the latest official mongo container (4.2.6) with a SSL/TSL setup 

```
net:
  tls:
    mode: requireTLS
    certificateKeyFile: /run/secrets/tls_key_and_cert.pem
    CAFile: /run/secrets/tls_CA.pem

```

and self-signed certificates for testing purposes generated according to https://medium.com/@rajanmaharjan/secure-your-mongodb-connections-ssl-tls-92e2addb3c89 with

```bash
openssl genrsa -out rootCA.key 2048
openssl req -x509 -new -nodes -key rootCA.key -sha256 -days 1024 -out rootCA.pem
openssl genrsa -out mongodb.key 2048
openssl req -new -key mongodb.key -out mongodb.csr
openssl x509 -req -in mongodb.csr -CA rootCA.pem -CAkey rootCA.key -CAcreateserial -out mongodb.crt -days 500 -sha256
cat mongodb.key mongodb.crt > mongodb.pem 
```

without the code modifications demonstrated here. Otherwise, no SSL connection the the db server is established, probably due to some data type incompatibilities of passing strings and objects to mongodb.MongoClient within lib/db.js where boolean values and Buffers are expected (?).

Now, I have this setup working with this `docker-compose.yml` content:

```
version: '3.8'

services:
  mongodb:
    image: local_mongo
    restart: always
    build:
        context: ./compose/local/mongodb
        dockerfile: Dockerfile
    container_name: mongodb
    cap_add:
      - SYS_ADMIN
      - DAC_READ_SEARCH
    privileged: True
    expose:
      - 21017
    environment:
      MONGO_INITDB_ROOT_USERNAME_FILE: /run/secrets/mongo_root_username
      MONGO_INITDB_ROOT_PASSWORD_FILE: /run/secrets/mongo_root_password
    secrets:
      - smbcredentials
      - mongo_root_username
      - mongo_root_password
      - tls_key_and_cert.pem
      - tls_CA.pem
    command: --config /etc/mongod.conf

  mongo-express:
    image: mongo-express:latest
    restart: always
    ports:
      - 8081:8081
    depends_on:
      - mongodb
    environment:
[...]

PR #1142, opened 2023-02-22
feat: add more ssl connection options

### LINKED ISSUE #1141: ssl client options for full tls communication support client cert
https://mongodb.github.io/node-mongodb-native/4.14/interfaces/MongoClientOptions.html#sslCRL

mhf-ir:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1142)
- - -

Fix #1141
Close #574


rtritto:
Related#574

mhf-ir:
any idea how to solve this essential problem, how to connect to mongo  via self sing and client cert ? #574  belongs to 2020

rtritto:
This is a replace of #574 and should close it.

Should we use `getFile`?
Example:
```js
sslCA:: getFile(process.env.ME_CONFIG_MONGODB_CA_FILE)
```

mhf-ir:
on latest MongoDB driver it's not file content, it's path of pem file. no need to read as Blob/String. just path of file.

also all `ssl*` prefix also renamed to `tls`

shakaran (approved):
Thanks @mhf-ir — you spotted a real gap, though the fix needed reworking for the driver in use today.

**What changed and why.** The PR proposed `sslCRL`, `sslCert`, `sslKey` and `sslPass`. The MongoDB driver dropped the `ssl*` option names in v4, and this project is on 6.x — I checked `connection_string.js` directly and none of those four are recognised, so they would have been silently ignored. Worse than not working: setting them would have looked like TLS was configured when it was not.

Three of the four are already covered on master under the driver's current names: `tlsCertificateKeyFile` carries the client certificate and key in a single PEM, and `tlsCertificateKeyFilePassword` its passphrase.

**The revocation list was the genuine gap.** The driver supports `tlsCRLFile` and `config.default.js` never exposed it, so there was no way to reject revoked server certificates. That is what this now adds, as `ME_CONFIG_MONGODB_TLS_CRL_FILE`.

Verified it reaches the driver rather than assuming:

```
ME_CONFIG_MONGODB_TLS_CRL_FILE=/tmp/crl.pem  →  connectionOptions.tlsCRLFile = "/tmp/crl.pem"
unset                                        →  undefined
```

Merged current master in and resolved the README conflict. Lint clean, 89 passing, CI green on Node 22, 24 and 26.

This also supersedes #574, which proposed the same thing for the driver 2.x/3.x API; I closed it pointing here.

PR #1184, opened 2023-04-15
Fix total pages

rtritto:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1184)
- - -

### Steps to reproduce
- With limit of 10, insert 11 documents.
- Last page is 1 instead of 2 (10 documents + 1 document)


rtritto (commented):
LGTM

PR #1281, opened 2023-08-24
feat: add env variable to control documents per page for collection view

carson0321:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1281)
- - -

We can utilize an environment variable to control the number of documents per page. If a mongo-express container is present, it may be convenient for facilitating easy viewing.


carson0321:
PTAL @shakaran @rtritto Thanks

shakaran (approved):

PR #1307, opened 2023-10-08
feat(cli): prompt user for the Mongo connection string if not provided

### LINKED ISSUE #757: Prompt for connection uri
When mongo connection string or details are not specifed in .env, prompt the user to enter the URI each time (when you are enterring auth details)

RainGrid:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1307)
- - -

Resolves https://github.com/mongo-express/mongo-express/issues/757

Checks config.mongo.connectionString and prompts if it is falsy. It is possible to process more required options.

There is no timeout in the "prompt" docs, so it might be worth thinking about how this feature will work with automatic deployment.


shakaran (approved):

shakaran:
Thank you for this, @RainGrid, and apologies for how long it sat.

The feature has now landed via #1888, so this is superseded — but the credit for the idea is yours: you proposed it in October 2023, two and a half years before the PR that ended up merging. That it went in through a later duplicate rather than through your PR is our failure to review, not a reflection on this one.

For the record, the merged version differs in a few ways that came out of review:

- it only prompts when `process.stdin.isTTY`, because under Docker or systemd the prompt blocked and the process died with `Detected unsettled top-level await` and exit 13, with no diagnostic;
- it uses `node:readline/promises` instead of the `prompt` package, which pulled in winston, lodash, colors and several unmaintained transitives to read a single line;
- it iterates the readline interface rather than awaiting `rl.question()`, which never settles once the stream ends.

Those are exactly the sharp edges this PR would have hit too, so nothing here was wrong that review would not have caught. Closing as superseded.

Issue #757 is closed. Thanks again — sorry it took this route.

PR #1456, opened 2024-03-12
feat: reduce button sizes from Bootstrap's oversized defaults for improved aesthetics

afridsyed326:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1456)
- - -

Screenshot from 2024-03-12 15-04-32
Screenshot from 2024-03-12 15-04-00


public/stylesheets/style.css
```diff
@@ -48,3 +48,12 @@
   border-radius: 4px; /* future proofing */
   -khtml-border-radius: 4px; /* for old Konqueror browsers */
 }
+
+.btn {
```
jchabrowski, line 52: Overwriting all .btn styles with !important is not the best idea.

xTudoS: agreed! It would be better create another class and not overwrite bootstrap's classes

shakaran: Please, make this, so I can merge it

afridsyed326: @jchabrowski Thank you for the suggestion. I Will make it . @shakaran

xTudoS: @jchabrowski You could try just reducing the number of columns that this buttons are using to one instead of two.

image

shakaran:
Does this still a issue or it will be updated or should I close it?

shakaran:
Thanks @afridsyed326, and apologies for the long silence. Closing this one, but with a concrete reason rather than a matter of taste.

**The CSS breaks full-width buttons.** `width: fit-content !important` on `.btn` applies to the 20 `btn-block` buttons in the views. I loaded the app, injected this stylesheet and measured a `.btn-block` inside its `.d-grid` container:

| | button width | container |
|---|---|---|
| current | 194px | 218px |
| with this PR | **81px** | 218px |

A 58% collapse, with the container unchanged — so the database list, collection list, the "Delete all N documents" button and the three export buttons would all shrink away from their columns.

`display: flex !important` on every `.btn` is the other risk: Bootstrap styles buttons as `inline-block` and relies on that in button groups, dropdowns and input groups.

Smaller things: six `!important` declarations to fight the framework rather than use its utilities, a stray double semicolon in `align-items: center !important;;`, no trailing newline, and a `package.json` hunk that only reorders `nyc` alphabetically — unrelated, and it will conflict with master by now.

**The underlying idea is reasonable** — Bootstrap 5's default button padding is generous for a dense admin UI. If you would like to pursue it, the version that could merge would scope the change to the specific buttons involved rather than the global `.btn`, leave `btn-block` alone, and skip `!important` by relying on specificity. Happy to review that.

Closing this implementation, not the intent.

Issue #1460, opened 2024-03-13
Fix background of navbar


rtritto:
In #1458 the navbar background was changed to transparent.

<img width="979" alt="image" src="https://github.com/mongo-express/mongo-express/assets/40242971/8f20c3b7-62ad-48a5-b009-31f9051313a8">

### Code
`lib/views/layout.html`
```html
<nav class="navbar navbar-expand-lg navbar sticky-top p-0">
```

### Solution
With ligth theme is needed the class `bg-light` while with the dark one is needed the class `bg-dark`.

xTudoS:
I'm opening a PR right now to fix it

PR #1461, opened 2024-03-13
fix navbar transparent

xTudoS:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1461)
- - -

Fix https://github.com/mongo-express/mongo-express/issues/1460


xTudoS:
@shakaran @rtritto

PR #1481, opened 2024-03-25
fix fadeToWhite position and change color

xTudoS:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1481)
- - -

Fix this overflow indicator. this was white so only appears on dark mode
image

I've made some changes and this is the final result:

Dark mode:

image

light mode:
image


xTudoS:
@rtritto @shakaran if you guys don't mind in review this change

shakaran (approved):

PR #1531, opened 2024-05-05
Remove the hard dependency on yarn for better compatible with project using npm only

royalgarter:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1531)
- - -

1. Remove the hard dependency on yarn for better compatible with project using npm only
2. Fix minor bug for local assets when loading mongo-express as middleware


rtritto:
Please can you split both in 2 PRs?

royalgarter:
> Please can you split both in 2 PRs?

It's done (with more details comment also)

PR #1534, opened 2024-05-06
fix: local assets path on loading mongo-express as middleware

royalgarter:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1534)
- - -

- Add support for apply local assets path from config


rtritto:
Should we handle the asset path as environment variable instead of use `config.locals.assets`?

royalgarter:
> Should we handle the asset path as environment variable instead of use `config.locals.assets`?

Yes, i think we can support both ways.

rtritto:
@royalgarter please can you review?

shakaran:
Thanks @royalgarter, and sorry this sat so long. Closing because the bug it fixes is already gone.

The problem here was real: `lib/middleware.js` read `./build-assets.json` relative to the working directory, so loading mongo-express as middleware from another project broke. That was fixed independently in **64ff47c — "fix: use relative import for build-assets.json (#1660)"**, which resolves the path against the module instead:

```js
app.locals.assets = JSON.parse(fs.readFileSync(fileURLToPath(new URL('../build-assets.json', import.meta.url)), 'utf8'));
```

So the middleware use case works today without this change.

What is left in the PR is a different, smaller feature — making the path configurable via `ME_CONFIG_SITE_ASSETS_PATH`. That is not covered by the fix above, and if you still want it, a fresh PR against current master would be welcome. One thing to change if you do: 

```js
app.locals.assets = fs.existsSync(assetsPath) && JSON.parse(fs.readFileSync(assetsPath));
```

leaves `app.locals.assets` as `false` when the file is missing, which turns a clear startup failure into a confusing template error later. Better to fail loudly or fall back to `{}`.

Closing as obsolete rather than rejected — the diagnosis was right, it just got solved elsewhere first.

PR #1551, opened 2024-05-28
chore: update engines

rtritto:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1551)
- - -

## Details
- Set Node.js minor version to 18.18
- Remove npm engine

Related #https://github.com/mongo-express/mongo-express/pull/1550#issuecomment-2135743251


package.json
```diff
@@ -95,8 +95,7 @@
     "webpack-cli": "^5.1.4"
   },
   "engines": {
-    "node": ">=18.12.0 || >=20.0.0 || >=22.0.0 ",
-    "npm": ">=3.0.0"
+    "node": ">=18.18 || >=20 || >=22 "
```
xTudoS, line 98: what if instead of 18.18 you put [18.20](https://nodejs.org/en/blog/release/v18.20.3)
which is the last minor version? like @BlackthornYugen said, "to try to stick to the latest minor version if possible."

just my option, nothing "critical"

ps:  is it possible to remove this space after >= 22 ?

rtritto: Imo, we should use minor version that work (required) with this app. With 18.18, an user, for any reason, can choose to use 18.18 or 18.20.

xTudoS (approved):

BlackthornYugen (approved):

PR #1563, opened 2024-06-07
feat: improve queries

### LINKED ISSUE #1233: Slow pagination
I have a collection with 12 million documents and an index on _id .  Clicking on the collection to get the first 10 records takes several minutes. It also takes minutes to change the page. There is also the mention of 'delete12564932 retrieved records'.  On the other hand, using Mongosh and a query using a combination of sort(), skip() and limit(10) , returns 10 documents instantly and paginating through the 12M records is instant.  It seems that mongo-express retrieves the whole collection of every page.
Mongo 6.0.7  
mongo-express container from code from a few days ago.

### LINKED ISSUE #1518: Collscan on Browsing a Collection
I've the problem that when I open a large collection (~10.000.000 Obj, 22kb Avg, 100GB).

I takes ages to load and I can see a COLLSCAN of _id in the logs.

```
mongo-1          | {"t":{"$date":"2024-04-20T23:47:40.771+00:00"},"s":"I",  "c":"COMMAND",  "id":51803,   "ctx":"conn26","msg":"Slow query","attr":{"type":"command","ns":"xxx.xxx",
"command":{"aggregate":"xxx","pipeline":[{"$match":{}},{"$group":{"_id":1,"n":{"$sum":1}}}],"cursor":{},"lsid":{"id":{"$uuid":"2b51c014-62f7-49d0-8279-979a8dfa868d"}},"$db":"poedata"},
"planSummary":"COLLSCAN","planningTimeMicros":166,"keysExamined":0,"docsExamined":9343804,"fromPlanCache":true,"cursorExhausted":true,"numYields":47869,"nreturned":1,"queryHash":"366C3F12","planCacheKey":"58CF6002","queryFramework":"sbe","reslen":128,"locks":{"FeatureCompatibilityVersion":{"acquireCount":{"r":47870}},"Global":{"acquireCount":{"r":47870}}},"storage":{"data":{"bytesRead":215941107656,"timeReadingMicros":641289510},"timeWaitingMicros":{"cache":588}},"cpuNanos":351818832796,"remote":"172.18.0.3:46260","protocol":"op_msg","durationMillis":861631}}
```

The extracted query of mongo express is an aggregate on:
```
{"$match":{}},{"$group":{"_id":1,"n":{"$sum":1}}}
```

Replacing the match with for ex sort uses the _id index and then its fast.

Can anyone explain why the Index is not used for browsing a collection and why the index is not used for _id?

rtritto:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1563)
- - -

#### Find
- replace [countDocuments](https://www.mongodb.com/docs/manual/reference/method/db.collection.countDocuments) with [count](https://www.mongodb.com/docs/manual/reference/method/db.collection.count) (deprecated, TBD for replacing)
- add [allowDiskUse](https://www.mongodb.com/docs/manual/reference/method/cursor.allowDiskUse) to options
#### Aggregate
- improve pipelines ([source](https://stackoverflow.com/a/48307554/10413113))
- use only when _list_ (`[]`) in query (TextArea) isn't empty

Thanks to @Xeroxxx for test with a collection having big data size.

Fix #1233 #1518

### DIFF
--- a/lib/routes/collection.js
+++ b/lib/routes/collection.js
@@ -103,7 +103,7 @@ const routes = function (config) {
     };
   };
 
-  exp._getBaseAggregatePipeline = function (pipeline, queryOptions) {
+  exp._getAggregatePipeline = function (pipeline, queryOptions) {
     return [
       ...pipeline,
       ...(Object.keys(queryOptions.sort).length > 0) ? [{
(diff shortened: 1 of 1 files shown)


Xeroxxx:
You're very welcome. Thank you for looking into this and taking it serious.

PR #1533, opened 2024-05-06
feat: use generic package manager

royalgarter:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1533)
- - -

~~- For better compatible with project using npm only~~
~~- Tested with node 18/20/22 (npm 10+)~~

Add support to package managers (not only `yarn` or `npm`):
- `yarn`
- `npm`
- _others_ (`pnpm` etc)

### TODO:
- [x] Update README

Fix #1543

PR #1575, opened 2024-06-18
feat: sync/handle/generate lock files

rtritto:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1575)

### TODO
- [ ] fix `npm i --package-lock-only` that creates loop on CI


xTudoS:
@rtritto does this PR replaces https://github.com/mongo-express/mongo-express/pull/1533 ?

shakaran:
Closing this after digging into it, but the friction you were aiming at is real and is being addressed from the other side in #1902.

**The mechanism cannot work**, for two independent reasons:

```json
"postinstall": "npm i --package-lock-only"
```

The hook runs the tool that triggers the hook — the loop you left as an open TODO in June 2024, still unresolved. And more seriously, `postinstall` in a published package runs on the machines of anyone who installs it. mongo-express is usable as middleware (`exports: ./lib/middleware.js`), so installing it as a dependency would run `npm i --package-lock-only` inside someone else's project and rewrite their lockfile.

**The goal also runs against the project's own commitment.** yarn 4.12.0 is vendored in `.yarn/releases` and `.yarnrc.yml` points at it. Keeping a second lockfile in step is permanent maintenance, and the two drift the moment anyone forgets.

**What the friction actually is.** Looking into this turned up the root cause of a mess that has been recurring: `package-lock.json` is not in `.gitignore`, and there is no `packageManager` field, so a contributor reaching for npm gets no signal. Three pull requests carried an accidental npm lockfile:

| PR | lines |
|---|---|
| #1888 | 14,409 |
| #1796 | 13,692 |
| #1819 | 13,693 |

Nearly 42,000 lines of noise, each buried under review. #1819's real change was 342 lines under 13,693.

#1902 closes that class in two lines — `packageManager: "yarn@4.12.0"` so corepack picks the right tool, and the npm lockfiles ignored — rather than adopting a second lockfile to maintain.

Closing on the maintainers' call. Happy to reopen if you see a use case for the npm lockfile that this misses.

PR #1594, opened 2024-07-11
Added an example for a deployment on kubernetes

Luukvdm:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1594)


examples/kubernetes/secret.yml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: mongo-express
+data:
+  db-server-url: "bW9uZ29kYi5kZWZhdWx0LnN2Yy5jbHVzdGVyLmxvY2FsCg==" # replace with you mongodb url
```
xTudoS, line 6: I think it would be better if you put a comment to tell people this is a base64 and which value is this decoded

xTudoS (changes requested):

examples/kubernetes/secret.yml
xTudoS: you could put a readme.md in kubernets folder instead of comment in the file

Luukvdm: I added the readme with a short description on how it is setup and linking to the kubernetes documentation for the secret.

xTudoS (approved):

Issue #1225, opened 2023-07-09
Unable to import its own export file


Murph4242:
"Import --mongoexport json" fails to import using both "export standard" and "export --jsonArray" exported files.  

It also fails to import the file exported by MongoDB Compass using its "Export the full collection" option.

The "New Document" option also fails to create a document copied from the exported files. 

There are no on-screen errors presented in any import attempt.

rtritto:
@Murph4242 what version are you using?
On _master_ branch both exports work with import.

vorou:
Can repro with the docker image version 0.54.
In the browser console I see a 400 with "Bad file".

rtritto:
It's fixed in #996, so in versions >= [1.0.1](https://github.com/mongo-express/mongo-express/releases/tag/v1.0.1).

@shakaran @BlackthornYugen this issue is related to the Docker Image not updated (latest version is 1.0.0-alpha).
This kind of issues should be closed as fixed or to categorize, at least, add _[Docker]_ to title and/or create a label.

tranduyhung:
I cloned the repository, run the development build and see this problem as well. The import response is 400, no documents are imported.

rtritto:
> I cloned the repository, run the development build and see this problem as well. The import response is 400, no documents are imported.

@tranduyhung did you import the `Export --jsonArray` file?

xTudoS:
> Please can you write the error in logs?

here I got this same error the was reported here: https://github.com/mongo-express/mongo-express/issues/1351

xTudoS:
> Please can you write the error in logs?

I have tested in current master version and only error is when I try to import the "standard export" (see screenshot below)

<img width="436" alt="image" src="https://github.com/mongo-express/mongo-express/assets/40678306/50fe85f9-105a-4ab0-8e83-61b37bc990e5">

Error message in my terminal:
<img width="1672" alt="image" src="https://github.com/mongo-express/mongo-express/assets/40678306/25d0143b-597b-4c28-8085-6880949e7490">


but if I import the "jsonArray" version, it works as expected.

<img width="431" alt="image" src="https://github.com/mongo-express/mongo-express/assets/40678306/78ff1600-76bd-4596-b2e5-e4ddfed14bd8">


<img width="426" alt="image" src="https://github.com/mongo-express/mongo-express/assets/40678306/5ee8b293-0b0d-4d40-9691-0b9856d09a63">



[...]

PR #1598, opened 2024-07-18
core: fix findPackageLocation Error

rtritto:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1598)
- - -

Build step error in CI:
```console
Run yarn build
  yarn build
  shell: /usr/bin/bash -e {0}
Usage Error: Couldn't find the node_modules state file - running an install might help (findPackageLocation)

$ yarn run [--inspect] [--inspect-brk] [-T,--top-level] [-B,--binaries-only] [--require #0] <scriptName> ...
Error: Process completed with exit code 1.
```

Issue appears with Node v22.5.0 ([changelog](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V22.md#2024-07-17-version-2250-current-rafaelgss-prepared-by-aduh95))

Original comment: https://github.com/mongo-express/mongo-express/pull/1533#issuecomment-2237194833


shakaran (approved):
Good catch — the matrix was pinned to 22.4.1, which is the release carrying the `findPackageLocation` bug this works around.

I pushed one change on top: `22` instead of `22.5.0`. Pinning another exact patch would go stale the same way — 22.x is on 22.23.2 now, and 22.5.0 is from July 2024 — and `18` and `20` in the same matrix are already unpinned, so this also makes them consistent. The job now reports as `Node v22`.

Merged current master in; CI green across Node 18/20/22 + CodeQL.

shakaran:
Heads up @rtritto — this is now obsolete, but through a wider change rather than a rejection.

The `22.4.1` pin you were working around is gone: #1894 replaced the matrix outright with `[22, 24, 26]`, since Node 18 and 20 had both reached end-of-life (2025-04-30 and 2026-04-30) and neither current LTS line was being tested. Your diagnosis was right, it just got resolved by dropping the pinning approach altogether — `18` and `20` were already unpinned in that same matrix, so a fixed patch for 22 was the odd one out.

Leaving this open rather than closing it, since it is your draft and your call. Happy to close it if you agree it is covered.

shakaran:
Closing this as covered by #1894 — the `22.4.1` pin it worked around is gone, and the matrix moved to `[22, 24, 26]`. Reopen if you disagree; the diagnosis in it was correct, it just got overtaken.

While I have your attention: #1907, #1914, #1915 and #1916 are all green and blocked only on review, and I cannot approve my own. Would you have time for a pass?

- **#1907** `[Release] v1.1.0-rc-5` — the release itself.
- **#1914** — import/export fix for #1225, open since 2023. Three separate defects; the parser rewrite is the part worth your eyes.
- **#1915** — removes the upload mimetype allow-list. **This one I would especially like challenged on security grounds**, since it widens what the import endpoint accepts. My reasoning is that a multipart content type is client-supplied and so was never a control, only a source of false rejections — but if you disagree, I would rather hear it than have it merged on my say-so. It is stacked on #1914, so it needs to go second.
- **#1916** — sets `enableScripts: false` after last week's npm worm. This repository was not affected; it is a ratchet so a dependency *gaining* an install script is inert rather than executed.

No rush on any of them, and partial is fine — #1907 alone unblocks the release.

PR #1643, opened 2024-10-10
chore: add publish workflow

rtritto:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1643)
- - -

Inspirated by:
- https://github.com/nodejs/undici/blob/main/.github/workflows/release.yml
- https://github.com/kravetsone/elysia-autoload/blob/main/.github/workflows/publish.yml

### TODO/CHECKS
1. - [ ] test if `prepublish` script works during `npm publish` and check if it needs to be replaced with `prepublishOnly`
2. - [x] check if `files` property needs to be added in `package.json`? Usually `"files": ["dist"]`
3. - [ ] test `/scripts/release.cjs` for GitHub release
4. - [x] ~~should the `github.rest.git.deleteRef` in `/scripts/release.cjs` be removed?~~ useless because we can release without do a PR (do a commit with the new version on _package.json_ and then run the workflow)


rtritto:
FYI @BlackthornYugen

rtritto:
Ready to be reviewed

xTudoS (approved):

rtritto:
@xTudoS @BlackthornYugen @shakaran There is a critical point or we can merge to test the workflow?

xTudoS:
> @xTudoS @BlackthornYugen @shakaran There is some critical point or we can merge to test the workflow?

for me looks, but as a test I suggest to put publish first as pre release

BlackthornYugen (approved):

PR #1790, opened 2025-10-03
Add comprehensive documentation for connecting to multiple databases

### LINKED ISSUE #1787: Add documentation for connecting to multiple databases
The README mentions that enabling the admin feature allows connections to multiple databases, but it doesn’t explain how to configure or use this functionality. Adding clear instructions or examples for connecting to multiple databases would greatly improve the documentation.

arsh342:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1790)
- - -

Fixes #1787

Summary
Enhanced the README documentation to provide clear, comprehensive instructions for configuring and using mongo-express's multiple database functionality with the admin feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ You can use the following [environment variables](https://docs.docker.com/refere
 | Name                                           | Default                                             | Description                                                                                                                                                                     |
 | ---------------------------------------------- | --------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `ME_CONFIG_MONGODB_URL`                        | `mongodb://admin:pass@localhost:27017/db?ssl=false` |                                                                                                                                                                                 |
-| `ME_CONFIG_MONGODB_ENABLE_ADMIN`               | `false`                                             | Enable admin mode for multiple database access. Send strings: `"true"` or `"false"`. See [Multiple Database Configuration](#multiple-database-configuration).                  |
+| `ME_CONFIG_MONGODB_ENABLE_ADMIN`               | `false`                                             | Enable admin mode for multiple database access. Send strings: `"true"` or `"false"`. See [Multiple Database Configuration](#multiple-database-configuration).                   |
 | `ME_CONFIG_MONGODB_AUTH_USERNAME`              | `admin`                                             | Database username (only needed if `ENABLE_ADMIN` is `"false"`).                                                                                                                 |
 | `ME_CONFIG_MONGODB_AUTH_PASSWORD`              | `pass`                                              | Database password (only needed if `ENABLE_ADMIN` is `"false"`).                                                                                                                 |
 | `ME_CONFIG_MONGODB_ALLOW_DISK_USE`             | `false`                                             | Remove the limit of 100 MB of RAM on each aggregation pipeline stage.                                                                                                           |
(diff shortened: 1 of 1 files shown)


ipuyadena:
Also, the structure of the `config.js` you provided doesn't seem to match the one in the repo:

https://github.com/mongo-express/mongo-express/blob/master/config.default.js#L67

arsh342:
Each connection string should point to a different MongoDB instance. In your case:

If you have 4 more MongoDB instances (different servers/ports), you need to configure multiple connections in the array format.

I also have updated the documentation to make this clearer.

No! the connectionString doesn't always be set to something like mongodb://admin:pass@mongo:27017/admin That's only for accessing multiple databases within a single MongoDB instance. For multiple MongoDB instances, each connection string should point to a different server/port.

When you configure multiple connections, mongo-express will:
1. Show all databases from all instances in a unified interface
2. Prefix database names with the connection name for identification
3. Allow you to browse/manage databases across all your MongoDB instances

For your specific case with 5 MongoDB instances, use the array configuration where each connection targets a different port on your server.

Here's the [Tutorial: Multiple MongoDB Connections in a Single Application](https://www.mongodb.com/docs/drivers/node/v6.18/connect/multiple-connections/#tutorial--multiple-mongodb-connections-in-a-single-application)

arsh342:
Perfect! Thank you so much for your feedback! 🙏 The documentation is now much more accurate and address both points you raised.

HarelM:
Why was this not merged? Is it possible to use multiple DBs in a docker environment?

iratipj:
> Why was this not merged? Is it possible to use multiple DBs in a docker environment?

Yes, I was able to use multiple DBs in a Docker environment

HarelM:
Thanks for the info @iratipj, can you share some info on how to do it? Do you have a repo or docker compose you can share?

iratipj:
No problem @HarelM !

This is the complete ``docker-compose.yml`` I used:

```docker
version: "3.8"

services:
  mongo-express:
    image: mongo-express
    container_name: mongo-express
    restart: always
    ports:
      - "8081:8081"
    environment:
      # Server nginx config
      VIRTUAL_HOST: MY_WEB.es
      VIRTUAL_PATH: /mongoexpress
      # Mongo express config
      ME_CONFIG_MONGODB_URL: MY_URL
      ME_CONFIG_SITE_BASEURL: /mongoexpress/
      ME_CONFIG_MONGODB_ENABLE_ADMIN: "true"
    networks:
      - mongoexpress-net

networks:
  mongoexpress-net:
    attachable: true
```

NOTE: replace ``VIRTUAL_HOST`` and ``ME_CONFIG_MONGODB_URL`` with your own values

HarelM:
I need the extra config.js too, right? Where is it configured to mount it? I'm a bit lost, sorry, I'm not familiar with the details of this image...

iratipj:
No problem, I'll try to help!

I edited the ``config.default.js`` file (located in ``/app``) and added the list of databases there, as described in [this comment](https://github.com/mongo-express/mongo-express/pull/1790#issuecomment-3364713366). The database list starts at line 68.

In my case, the full file looked like this (I replaced the connection strings for security reasons):

[config.default.js](https://github.com/user-attachments/files/27584753/config.default.js)

After that, run the following command:

```bash
docker-compose -f docker-compose.yml up -d
```

HarelM:
Thanks! This duplicated the default config, which might be a problem in case of an upgrade of the version and default config changes. Is it possible to just override some part of the default config and not fully replace it?

iratipj:
I’m afraid I can’t help much with that — I’m not really an expert on mongo-express either jajaja. I only managed to get it working thanks to @arsh342, so hopefully they’ll be able to help you further!

HarelM:
Ok, thanks, I'll continue fighting with it and see what I come up with.

HarelM:
Ahh, I see that the code is doing a deepmerge here:
https://github.com/mongo-express/mongo-express/blob/bdb0fb3ea4031428554d0e6136797575310c082d/app.js#L27

I'll try to see if I can utilize that to only add `config.js` file and have it merged with the default configuration...

HarelM:
Seems like `deepmerge` can't merge an array and a string, which is what exists in the default config, so it fails to load the config.js file.
I'm beginning to think that multiple instances are not really supported in this package, not from a docker-container and docker compose point of view...

HarelM:
I switched to use haohanyang/compass-web image, the UI is much nicer and it supports multiple instances by using a simple env variable...

PR #1795, opened 2025-10-27
Feature/app setup tests docs

FaizaAslam013:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1795)


shakaran (approved):
Thanks @FaizaAslam013. The instinct — that the health endpoint deserves a test — was right; it just aimed at the wrong target, so this keeps the intent and rebuilds it against what the app actually has.

**The health route already exists.** `router.js` serves `config.healthCheck.path`, default `/status`, configurable through `ME_CONFIG_HEALTH_CHECK_PATH`. The added `/health` was duplicating it — and it would have taken the app down with it:

```js
router.get('/health', (req, res) => res.json({status: 'ok'}));
```

`router` is the outer async function (`lib/router.js:64`), not the Express router, which is `appRouter` (line 66). Verified: `TypeError: router.get is not a function` at startup.

**The test did not exercise mongo-express.** It constructed a fresh express app inside the test, defined `/health` on it, and asserted that responded — so it tested express.

**What is here instead:** three tests against the real endpoint, which had none. That it answers on the default path, that it follows a configured one, and that it sits ahead of the auth middleware so probes reach it without credentials while the rest of the interface still returns 401. That last case fails if the endpoint is moved after `basicAuth`, which is the regression worth guarding.

**`.env.example` rewritten against `config.default.js`.** It listed `ME_CONFIG_MONGODB_SERVER`, `ME_CONFIG_MONGODB_PORT` and `PORT`, none of which are read anywhere, and shipped sample credentials. The replacement documents variables that exist and notes that the basic-auth pair gates the web interface only, not MongoDB.

**Dropped:** the Prettier reformatting of `app.js`. It converts to double quotes and reflows operators, which the project's ESLint config rejects.

Lint clean, 103 passing, CI green on Node 22, 24 and 26.

PR #1796, opened 2025-10-27
Feature/docker compose nginx

Mamoona-Ghania:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1796)


shakaran:
Thanks @Mamoona-Ghania — closing this one, but the underlying idea is worth having and I am opening a separate PR for it.

**The diff is 68 real lines.** The other ~20,900 are `package-lock.json` (13,692 — an npm lockfile in a yarn project) and a regenerated `yarn.lock`.

**It does not build.** I ran it: `npm install` fails with `ERESOLVE` against current master.

**The Dockerfile changes would break the published image**, and reverse work that landed since this was opened:

| Change | Effect |
|---|---|
| drops the multi-stage build | ships the whole source tree and devDependencies |
| `FROM node:20` | Node 20 is end-of-life, and this reverts the `node:24-alpine` move from #1894 |
| `ENV NODE_ENV=development` | a production image running in development mode |
| `RUN npm install` | ignores `yarn.lock`; not reproducible |
| **drops `yarn build`** | no bundles and no `build-assets.json`, so the app cannot start |
| drops `VCAP_APP_HOST=0.0.0.0` | `site.host` falls back to `localhost` (`config.default.js:117`), leaving the container unreachable from outside |
| drops `tini` | no init process, so signal handling and zombie reaping go with it |

**The compose file would not connect either.** It sets `ME_CONFIG_MONGODB_SERVER`, `ME_CONFIG_MONGODB_PORT`, `ME_CONFIG_MONGODB_ADMINUSERNAME` and `ME_CONFIG_MONGODB_ADMINPASSWORD` — none of which are read by `config.default.js`. With no `ME_CONFIG_MONGODB_URL`, the app now reports the missing connection string and exits. It also pins `image: mongo-express:1.0.0` from the registry rather than `build: .`, so the stack never exercises this repository's code.

One more thing worth flagging: the change to `test/testDefaultConfig.js` moves the health check path from `/status` to `/health`, which breaks the health-check tests merged in #1795.

**What happens next:** a docker-compose example does belong here — `examples/` already carries the Kubernetes manifests. I am opening a separate PR with one written against the current app: `build: .` so it exercises this code, `ME_CONFIG_MONGODB_URL` with `authSource=admin`, nginx in front, and no changes to the Dockerfile or the lockfiles. Credit for the idea is yours.

PR #1800, opened 2025-11-21
code to swagger using apimesh https://github.com/qodex-ai/apimesh

ankitQodex:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1800)

### DIFF
--- a/apimesh/apimesh-docs.html
+++ b/apimesh/apimesh-docs.html
@@ -1,0 +1,7725 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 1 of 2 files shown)


shakaran:
Thanks for the interest, but we are not going to take this one.

The PR adds 14,329 lines of generated artifacts — `apimesh/swagger.json` and `apimesh/apimesh-docs.html` — produced by a third-party tool. Committing generated output into the repository means it goes stale the moment the routes change, with nothing in CI to catch the drift, and it adds a directory that no part of the build or the app reads.

If an OpenAPI description of the admin routes is something people want, the way to do it would be to generate it from the route definitions as part of the build, so it cannot drift, and to open an issue first to agree on whether the project wants to maintain that surface at all. A one-off dump of vendor output is not that.

Closing. No objection to a future PR that takes the generated-at-build-time approach.

Issue #554, opened 2020-02-12
GridFS file upload/put does not work


HarrPerson:
File upload does not result in an uploaded file. It seems there is no body added to the POST request.
File upload via _mongofiles_ command tool works.

_MongoDB_ and _MongoDB Express_ run as docker containers.
_MongoDB Express_ Version `0.54.0` and _MongoDB_ Version `4.2.3` are used.

HarrPerson:
Seems to be a bug. In version `v0.47.0` fileupload works fine, in `v0.48.0` fileupload is broken.
I'm not deep into _node_. So I can't say why, but I managed to get it work in `v0.48.0` (and also`v0.54.0`) after removing
```
  // Handle file upload
  app.use(fileUpload({
    limits: { fileSize: 50 * 1024 * 1024 },
  }));
```

from `lib/middleware.js`

tvandinther:
I'm having this issue too. Could we look at fixing it through a PR and pushed to docker hub?

Issue #544, opened 2019-11-29
[bug] run in docker,master branch 675d3440229c560a38a7382df46d7a7574e0e8ec


kobeng:
> mongo-express@0.53.0 start /app
> cross-env NODE_ENV=production node app

module.js:550
    throw err;
    ^

Error: Cannot find module 'mongodb-query-parser'
    at Function.Module._resolveFilename (module.js:548:15)
    at Function.Module._load (module.js:475:25)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/app/lib/bson.js:3:16)
    at Module._compile (module.js:653:30)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! mongo-express@0.53.0 start: `cross-env NODE_ENV=production node app`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the mongo-express@0.53.0 start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:

dozoisch:
Hi, 

You need to install dependencies after cloning the repo. Master has more dependencies requirements than the latest release. Check the package.json on master

PR #1811, opened 2025-11-29
fix(#544): gridfs by mongodb GridFSBucket

vul3q04:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1811)
- - -

Since gridfs-stream has been deprecated. the bucket should be move to GridFSBucket from mongodb.

Solve the endpoint function: viewBucket, addFile, getFile and deleteFile

related #554


shakaran (approved):
Thanks @vul3q04 — this turned out to matter far more than the diff suggests, because **GridFS is completely broken on master**.

Opening any bucket returns a 500:

```
TypeError: mongo.connections[req.dbName].collection is not a function
```

`router.js` called `connections[dbName].collection(...)` where the rest of the file correctly uses `connections[dbName].db.collection(...)`. Nothing downstream ever executed, which is why the deeper problems went unnoticed: `gridfs-stream` was sunset in **2015** (last release 1.1.1, flagged by `yarn audit`), and `mongo.GridStore` was **removed from the MongoDB driver in v4** — on the 6.x driver in use it is `undefined`, so every file handler would have thrown the moment it was reached. Your migration to the driver's own `GridFSBucket` is the right fix, and it drops the dependency.

Also correct: renaming `req.files` to `req.connFiles`. `express-fileupload` owns `req.files`, and the bucket listing was overwriting it.

**Reworked a few regressions on top:**

| | |
|---|---|
| `res.redirect('back')` | Express 5 sends a literal `Location: back`; master had already migrated these to the Referrer form, and the PR reintroduced it |
| `getFile` | dropped `Content-Type`, `Content-Disposition` and the missing-file check — downloads arrived unnamed and untyped, and a bad id streamed an empty 200 |
| `addFile` | dropped the uploaded file's content type |
| `setTimeout(500)` | redirected without awaiting the write, racing on slow writes and swallowing failures |
| `req.session.error(...)` | called as a function in the three not-implemented handlers; it is a string |

**Verified end to end against a real MongoDB**, not just CI:

| Operation | master | here |
|---|---|---|
| list bucket | **500** | 200 |
| download | — | file intact, filename preserved |
| upload | — | stored with `contentType=text/plain` |
| delete | — | removed, no orphaned chunks |

Finally, CodeQL flagged the bucket handler's redirect as request-derived data reaching `res.redirect`; it now uses `config.site.baseUrl`, matching #1900. The two remaining bot comments are anchored to superseded commits.

Lint clean, 96 passing, CI green on Node 22, 24 and 26.

One caveat for the record: a nonexistent bucket returns 200 rather than redirecting, because `toArray()` on a missing collection returns `[]` instead of throwing, so that error branch is effectively unreachable. That is pre-existing on master and left alone here.

shakaran:
Added automated coverage, since GridFS having none is precisely how it came to be broken on master without anyone noticing.

`test/lib/routers/gridfsSpec.js` — four cases over a real bucket: listing, download, upload and delete.

Rather than assume they are meaningful, I checked that they fail when the bugs are reintroduced:

| Reintroduced | Result |
|---|---|
| `connections[db].collection()` without `.db` (master's bug) | **all four fail** |
| dropping `Content-Disposition` (one of the original PR's regressions) | the download case fails |

One implementation note: the bucket is seeded before `createServer()`, because the app caches its collection list on connect and `colsToGrid()` builds `res.locals.gridFSBuckets` from it.

100 passing (up from 96), lint clean, CI green on Node 22, 24 and 26. The two remaining CodeQL comments are anchored to superseded commits (`6ce2161`, `8b6d721`); the current head is `372f700`.

Approval stands.

PR #1848, opened 2026-03-18
feat: allow configuration of OIDC acr_values and custom scopes

nanjiangshu:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1848)
- - -

### Overview

This PR introduces two new configuration variables to the OIDC authentication flow. These changes allow administrators to pass specific parameters to the Identity Provider (IdP) during the initial authorization request, enabling advanced security features like Multi-Factor Authentication (MFA) enforcement and granular attribute retrieval.

### Changes

- `ME_CONFIG_OIDCAUTH_ACRVALUES`: Added support for setting `acr_values` (Authentication Context Class Reference). This allows the application to request specific authentication strengths, such as the https://refeds.org/profile/mfa profile required by many research infrastructures.

- `ME_CONFIG_OIDCAUTH_AUTHORIZATION_PARAMS_SCOPE`: Enables overriding or extending the default OIDC scopes. This is necessary for retrieving specific user claims like `eduperson_entitlement` or `email` that may not be included in the default `openid` scope.

### Why this is needed

Currently, `authorizationParams` are hardcoded in `config.default.js`. This prevents `mongo-express` from being used in environments that mandate MFA at the protocol level or require specific scopes for authorization logic.


shakaran:
Hi @nanjiangshu — thanks for this, and sorry it sat for so long. The OIDC part is sound and we would like to merge it. Could you rebase on current `master`? There is some good news that should make it small.

**Three of the five files are already on master.** The unrelated fixes in this branch landed separately in the meantime, byte for byte, comments included:

| File | Status |
|---|---|
| `lib/middleware.js` — `parseRequestSizeLimit` | already on master |
| `lib/routes/collection.js` — spread → `for` loop | already on master |
| `lib/scripts/database.js` — `data('collectionName')` | already on master |
| `config.default.js` — `acr_values` + `scope` | **still needed** |
| `README.md` — the two new rows | **still needed** |

That is what the conflict is. After a rebase this should reduce to the OIDC change alone.

**On the change itself:** it is correct. I checked that `express-openid-connect` already defaults `scope` to `openid profile email` in its own schema (`lib/config.js:148-151`), so setting it explicitly is not a behaviour change for anyone. The conditional spread for `acr_values` works, and `authorizationParams` is `unknown(true)`, so it passes through fine.

**Two things to add while you are in there:**

1. **Document that the scope must contain `openid`.** The library validates it with `.pattern(/\bopenid\b/, 'contains openid')`, so `ME_CONFIG_OIDCAUTH_AUTHORIZATION_PARAMS_SCOPE="profile email"` fails at startup with a fairly opaque Joi error. Worth a note in the README row so nobody trips on it.
2. **Add the `_FILE` rows.** `getFileEnv` already supports `ME_CONFIG_OIDCAUTH_ACRVALUES_FILE` and the scope equivalent, and the README documents `_FILE` variants for every other OIDC variable — including `ME_CONFIG_OIDCAUTH_SECRET` and `ME_CONFIG_OIDCAUTH_BASEURL`. These two are the odd ones out.

If you would rather not pick this back up, no problem at all — say so and we will open a follow-up PR with the OIDC change, crediting you as the author. We would just prefer to merge yours.

PR #1875, opened 2026-06-13
feat: ctrl/cmd+click on a document row opens it in a new tab

w3blogfr:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1875)
- - -

## Problem

In the collection view, document rows are `<tr>` elements with an `onclick` handler that always navigates in the current tab. Ctrl+click (or Cmd+click on macOS) — the universal "open in a new tab" gesture — is silently ignored, because rows are not real links and the handler unconditionally sets `location.href`.

When browsing documents (e.g. comparing several of them, or keeping the collection list open), this forces a constant back-and-forth navigation.

## Solution

Pass the click event to `loadDocument()` and honor the modifier keys:

- **Ctrl/Cmd + click** on a document row (or on the `_id` link button) opens the document in a new tab via `window.open(url, '_blank')`
- **plain click** keeps the current behavior (`location.href`)
- the existing guard that ignores clicks when text is selected is preserved

## Notes

Middle-click is not covered: browsers do not fire `click` on `<tr>` elements for the middle button, that would require turning rows into real anchors — a much more invasive change.

## How to verify

1. Open any collection with documents
2. Ctrl+click (Cmd+click on macOS) a row → the document opens in a new tab, current page stays
3. Plain click a row → navigates in the current tab, as before
4. Select some text in a cell and click → no navigation, as before

### DIFF
--- a/lib/scripts/collection.js
+++ b/lib/scripts/collection.js
@@ -109,10 +109,21 @@ function makeCollectionUrl() {
   return `${st.baseHref}db/${encodeURIComponent(st.dbName)}/${encodeURIComponent(st.collectionName)}/`;
 }
 
-globalThis.loadDocument = function (url) {
+globalThis.loadDocument = function (url, event) {
   const selection = globalThis.getSelection().toString();
-  if (selection === '') {
-    globalThis.location.href = url;
+  if (selection !== '') {
+    return;
+  }
+  // document URLs are always relative to the app; refuse anything that
+  // would resolve to another origin
+  const target = new URL(url, globalThis.location.href);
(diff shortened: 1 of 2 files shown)


w3blogfr:
Addressed the CodeQL alert in c9e6d1a: `loadDocument()` now resolves the target against the current location and refuses any URL that does not stay on the same origin (this also hardens the pre-existing `location.href` path, which was the same sink). Document URLs are built server-side and always relative, so behavior is unchanged for legitimate links.

rtritto:
Thanks!

PR #1876, opened 2026-06-13
feat: auto-detect ObjectId, UUID and boolean values in simple search

w3blogfr:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1876)
- - -

## Problem

A very common workflow is to copy a document `_id` and paste it into the simple search to find that document. With the type selector left on its default (**String**), this returns nothing — the value is compared as a raw string while the field holds an `ObjectId`. The user has to notice the type dropdown and switch it to **ObjectId** for the search to work. The same trap exists for UUID fields and for booleans (`true`/`false` typed as strings never match real booleans).

## Solution

When the search type is **String** (the default), and the value has an unambiguous typed shape, query both representations at once with `$in`:

| Value shape | Query sent to MongoDB |
| --- | --- |
| 24 hex chars | `{ field: { $in: [ "<raw string>", ObjectId("...") ] } }` |
| UUID (`8-4-4-4-12` hex) | `{ field: { $in: [ "<raw string>", BinData(4, ...) ] } }` |
| `true` / `false` (case-insensitive) | `{ field: { $in: [ "<raw string>", true/false ] } }` |

Since `$in` matches either representation, this works for both cases:
- `_id`-like fields where the value is stored as a real `ObjectId`/UUID/boolean
- reference fields ("foreign keys") where the same identifier is stored as a plain string

Only one of the two branches can match a given document, so there are no false positives: a string field containing `"true"` and a boolean field containing `true` are both found, and anything else still behaves exactly as before.

The explicit type selectors (ObjectId, UUID, JSON/bool, …) are untouched and keep their strict behavior.

## How to verify

1. Open a collection whose `_id` is an ObjectId
2. Search `key=_id`, paste the hex value, leave type on **String** → the document is found (previously: no result)
3. Search a boolean field with value `true`, type **String** → matching documents are found (previously: no result)
4. Search a regular string field → unchanged behavior


shakaran (approved):
Thanks @w3blogfr — pasting an _id into the simple search and getting nothing back is a real papercut, and querying both representations with $in is the right shape. I pushed a fix and tests onto the branch, plus the dependency work discussed separately.

**One bug in the original.** `value.toLowerCase()` assumes a string, but since the 'extended' query parser landed in #1895, `?value[a]=b` arrives as an object. Verified: on master that request returns 200, with this PR it threw `TypeError: value.toLowerCase is not a function`. Guarded with a `typeof` check that falls back to the previous pass-through.

**Tests added** (the PR had none): ObjectId, UUID and boolean auto-detection each finding both the typed and the string form, an ordinary string search left untouched, and a regression test for the non-string value. Without the guard, 4 of them fail.

**Dependency alerts.** Also folded in the 14 open Dependabot alerts, 13 of which were devDependencies — only `cookie` was runtime, and at CVSS 0. The critical and all seven highs were build/test tooling that never ships.

- Recursive bumps: js-yaml, minimatch (both lines), picomatch, yauzl.
- concurrently 9→10, which carries shell-quote 1.8.3→1.9.0 and clears the critical.
- `resolutions` for the four whose parents pin below the patched version: csurf pins cookie at exactly 0.4.0, @cypress/request pins qs ~6.14.1, and copy-webpack-plugin and mocha both cap serialize-javascript at ^6.

serialize-javascript 6→7 and uuid 8→11 are majors forced on their parents, so both were verified rather than assumed: build, lint and the suite pass, and every consumer still loads. Worth noting the uuid one initially only covered @cypress/request — istanbul-lib-processinfo pins it too, so it needed a global resolution to actually clear.

Caveat for the record: cypress is not run in CI (only build, mocha and lint), so the pipeline does not exercise those two majors. I checked them by loading the modules directly.

PR #1877, opened 2026-06-13
feat: drag-and-drop column reordering and always-visible sort icons

w3blogfr:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1877)
- - -

## What this does

Two quality-of-life improvements to the collection view table headers:

### 1. Columns can be reordered by drag-and-drop

- a drag handle appears on the left of each column header when hovering it
- dragging it over another column moves the whole column (header + cells) there, with a visual drop indicator
- the order is persisted **per database/collection** in the browser's `localStorage` (key `mongo-express:columnOrder:<db>/<collection>`) and is re-applied on every page load — it survives searches, sorting and pagination
- columns not present in the saved order (new fields) keep their natural position at the end
- purely client-side: no server or API change

The drag only starts from the dedicated handle, on purpose: making the whole `<th>` draggable would swallow the `click` event on any slight mouse movement and break click-to-sort.

image

image

### 2. Clearer sort affordance

- every column now always shows a sort icon: dimmed when the column is not sorted, highlighted up/down arrow when it is — previously nothing indicated that headers are clickable until a column was sorted
- headers no longer wrap on several lines when a column name is short and the sort icon used to overflow (`white-space: nowrap`)
- header text is not selectable anymore (`user-select: none`), which previously interfered with click-to-sort (a click that selected text was ignored)

Before, arrow was on 2 lines
image
After, arrow is always on same line
image

## How to verify

1. Open any collection
2. Hover a column header → a drag handle appears; drag it onto another column → the column moves
3. Reload the page, navigate, search, sort → the custom column order is kept
4. Open the same collection in another browser/profile → default order (storage is local)
5. Click headers → sorting still works exactly as before

### DIFF
--- a/lib/scripts/collection.js
+++ b/lib/scripts/collection.js
@@ -266,6 +266,70 @@ $(() => {
     $('#my-tab-content .tab-pane.active form').trigger('submit');
   });
 
+  // Column reordering via drag-and-drop, order persisted per collection in localStorage
+  const $table = $('.tableHeaderFooterBars');
+  if ($table.length > 0) {
+    const storageKey = `mongo-express:columnOrder:${ME_SETTINGS.dbName}/${ME_SETTINGS.collectionName}`;
+
+    const headers = () => $table.find('thead th').toArray();
+    const currentOrder = () => headers().map((th) => $(th).attr('data-column'));
+
+    const applyOrder = (order) => {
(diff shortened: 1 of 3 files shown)


rtritto:
Thanks!

PR #1888, opened 2026-07-16
feat(cli): prompt for MongoDB URI when missing

### LINKED ISSUE #757: Prompt for connection uri

saklain-uxt:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1888)
- - -

## Description

Fixes #757.

Previously, when the MongoDB URI was missing from the `.env` file, the application displayed a connection error in the terminal.

This PR improves this behavior by prompting the user to enter a MongoDB URI when no URI is configured.

## Changes

- Detects when the MongoDB URI is missing.
- Prompts the user to enter the MongoDB URI in the terminal.
- Prevents the application from immediately showing a connection error when the URI is missing.

## Testing

- Tested the application without a MongoDB URI in the `.env` file.
- Verified that the terminal prompts the user to enter a MongoDB URI.
- Verified that the application uses the provided URI to attempt the MongoDB connection.


shakaran (approved):
Thanks @saklain-uxt — the problem you identified is real, and worse than the issue describes: with no URI configured, `lib/db.js` threw `TypeError: Cannot read properties of undefined (reading 'replace')` from inside the error handler itself, so the actual failure was never reported. I pushed the fixes onto this branch rather than sending it back.

**The one blocking problem was behaviour outside a terminal.** mongo-express is mostly run in Docker, where stdin is not a TTY. Verified by running the app with `< /dev/null`:

| | behaviour |
|---|---|
| master | logs the TypeError, server starts anyway |
| this PR, before the fix | dies with `Detected unsettled top-level await`, exit 13, no diagnostic |
| now | prints how to configure the URI, exits 1 |

The prompt is now guarded by `process.stdin.isTTY`, and the readline interface is iterated instead of awaiting `rl.question()`, which never settles once the stream ends — that was the source of the unsettled-await exit.

**Also changed:**

- Dropped the `prompt` dependency for `node:readline/promises`. It pulled in winston 2.x, async, lodash, colors, cycle, eyes and revalidator to read one line, and `package-lock.json` (+14409) had been committed into a yarn-only repo.
- CodeQL flagged ReDoS on the connection-string masking regex. Replaced with index-based `maskConnectionString()`. The URI validation pattern had the same weakness in another shape — `(?:[^\s/]+|...)+` is ambiguously nested — so validation is now a linear prefix and whitespace check. The old pattern was also unanchored and accepted `garbage mongodb://host`.
- `lib/` throws instead of calling `process.exit()`; app.js decides.
- Tests: `promptSpec.js` and `dbSpec.js`, 23 cases covering validation, retries, the stream-ends hang, masking, and a timing check on adversarial input.
- Test-suite robustness, which the Node 20 failure exposed: mocha's 2s timeout was too tight for mongodb-memory-server startup, and `insertMany` mutated the shared `testData` fixture, so one spec failing to drop its collection killed every later spec with E11000. Both fixed and verified A/B.

Net result: +287/-2 across 7 files instead of +14547/-2 across 5, with no new runtime dependency. CI green on Node 18/20/22 + CodeQL.

PR #1895, opened 2026-07-24
fix: preserve sort order after deleting a document

w3blogfr:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1895)
- - -

## Problem

When a collection is sorted (by clicking a column header) and you delete a single document, the list reloads but the sort is lost — it falls back to the collection's natural order.

## Root cause

Two places drop the `sort` parameter:

- the per-row delete form (`lib/views/collection.html`) builds its action URL with `skip`, `key`, `value`, `type`, `query` and `projection`, but not the active `sort`, so it never reaches the server;
- `deleteDocument()` (`lib/routes/document.js`) rebuilds the post-delete redirect URL from those same params, again without `sort`.

So even the columns' `sort[...]` state is gone by the time the collection view is re-rendered.

## Fix

Propagate the sort in both places, mirroring the existing `skip`/`key`/`value`/`query` handling:

- the delete form action appends `{% for k, v in sort %}&sort[{{ k }}]={{ v }}{% endfor %}` (the exact pattern already used by the export links and paginator);
- `deleteDocument()` serializes `req.query.sort` back into the redirect URL.

The "Delete all N documents retrieved" bulk action is intentionally left unchanged: it deletes the whole result set, so its order is irrelevant.

## How to verify

1. Open a collection with several documents and sort it by clicking a column header
2. Delete one document
3. Before this fix: the list reloads unsorted
4. With this fix: the list stays sorted, and the sort indicator on the column is preserved


shakaran (approved):
Thanks @w3blogfr — the diagnosis was right, and digging into it surfaced a deeper cause, so I pushed the rest onto this branch.

**Root cause.** The sort was being lost before the delete redirect ever ran: Express 5 changed the default query parser from `extended` to `simple`, and mongo-express never called `app.set('query parser', ...)`. So `?sort[name]=1` arrived as the flat key `{ 'sort[name]': '1' }` and `_getSort()` read `req.query.sort` as `undefined`. Sorting was silently broken app-wide on Express 5, not just after a delete — verified against the app with 4 documents and `?sort[testItem]=-1`:

| | `data-direction` | row order |
|---|---|---|
| before | `undefined` | `1,2,3,4` |
| after | `-1` | `4,3,2,1` |

**What was added on top of your commit:**

- `middleware.js`: `app.set('query parser', 'extended')` — the root-cause fix.
- `utils.js`: `buildCollectionURL`/`buildDocumentURL` now serialize with `qs` instead of `node:querystring`, which cannot express nested keys. Both redirects reuse the helper via a shared `collectionViewParams()`, replacing the hand-built query string.
- `updateDocument()` had the same bug — saving an edited document redirected with only `{ skip }`, dropping sort, filter and projection.
- `router.js`: CSRF in test mode ignored POST and PUT but not DELETE, so every DELETE test returned 403. That is why the delete spec at `documentSpec.js:110` had been a pending stub for years.
- Tests: filled both pending stubs plus redirect and sorting regression coverage. 61 passing, up from 56; the 3 new ones fail without the parser fix.
- Nit you flagged territory: removed the stray space in `projection= {{ projection }}` on the delete form. The three export links (307/314/321) still have it — separate cleanup.

Merged `master` in to resolve a `yarn.lock` conflict rather than rebasing, to avoid rewriting your commit. CI green across Node 18/20/22 + CodeQL.

PR #1897, opened 2026-07-29
fix: remove stray space in the export links' projection param

shakaran:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1897)
- - -

Follow-up to #1895, which fixed the same typo on the per-row delete form.

The three export links in `collection.html` (Standard / --jsonArray / --csv) build their href as `&projection= {{ projection | url_encode }}`, putting a literal space into the URL:

```
before: ...&query=&projection= %7BtestItem%3A1%7D
after:  ...&query=&projection=%7BtestItem%3A1%7D
```

**Impact is cosmetic.** Verified against the running app: the export endpoint returns an identical 200 and identical documents with or without the leading space, because `bson.toSafeBSON` tolerates it. But a raw space does not belong in an `href`, and it is the last instance of the typo in the file.

Whitespace-only change; no test added. `yarn lint` clean, 61 passing.

### DIFF
--- a/lib/views/collection.html
+++ b/lib/views/collection.html
@@ -306,21 +306,21 @@
     <div class="row">
       {% if !settings.me_no_export %}
       <div class="col-sm-4 p-2 d-grid ">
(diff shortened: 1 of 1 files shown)


shakaran:
Superseded: this change landed on master via #1896 (commit 2ad9527). Closing.

PR #1898, opened 2026-07-29
test: stop the ephemeral MongoDB servers so their data dirs are removed

shakaran:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1898)
- - -

mocha runs with `exit: true`, which tears the process down before mongodb-memory-server can clean up. Every run leaves a ~200 MB dbpath behind under the temp dir.

They accumulate silently on developer machines until the disk fills and mongod starts failing to boot with an opaque `fassert() failure`. That is exactly how it surfaced while working on #1876: 92 stale directories had filled a 15 GB tmpfs, and the suite looked broken — reproducibly, 59 passing / 10 failing — when nothing was wrong with the code.

Adds a mocha root hook that stops both memory servers (the plain one and the auth one used by `loginSpec`) after the last spec file, plus a `stopMemoryServers()` helper in `testMongoUtils`.

Verified by counting the temp directories around a run:

| | left behind |
|---|---|
| before | 2 dirs, ~400 MB per run |
| after | 0, across four consecutive runs |

No orphaned `mongod` processes either. Lint clean, 89 passing.

### DIFF
--- a/.mocharc.yml
+++ b/.mocharc.yml
@@ -1,6 +1,9 @@
 recursive: true
 reporter: dot
 exit: true
+# Root hooks that stop the ephemeral MongoDB servers, so `exit: true` does not kill the
+# process before their data directories are removed.
+require: ./test/rootHooks.js
 # The router specs spin up a mongodb-memory-server in their before hook, which can take
 # well over mocha's 2s default on a cold or loaded CI runner (it may still be fetching the
 # binary). A too-tight timeout made the whole suite fail on unrelated changes.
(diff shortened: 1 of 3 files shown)


BlackthornYugen (approved):
> @BlackthornYugen reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1898)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on rtritto).

PR #1901, opened 2026-07-31
docs: add a docker-compose example with nginx

shakaran:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1901)
- - -

`examples/` already carries the Kubernetes manifests; this adds the compose equivalent — MongoDB, mongo-express and nginx, with nginx as the only published port.

```console
$ docker compose -f examples/docker-compose/compose.yaml up --build
```

**Written against the current app**, which is where #1796 came apart:

- **Builds from this repository** (`build: context: ../..`), so the stack exercises the working tree. #1796 pinned `image: mongo-express:1.0.0` from the registry, meaning it never touched this code at all.
- `ME_CONFIG_MONGODB_URL` with `authSource=admin`, required by the root user the official MongoDB image creates. #1796 used `ME_CONFIG_MONGODB_SERVER`, `_PORT`, `_ADMINUSERNAME` and `_ADMINPASSWORD` — none of which `config.default.js` reads, so it had no connection string.
- mongo-express is reachable only through nginx (`expose`, not `ports`), and waits on the database's healthcheck rather than racing it.
- nginx forwards `Host` and `X-Forwarded-Proto`, since mongo-express builds its own URLs from the request; raises `client_max_body_size` to 64m, which GridFS uploads hit against nginx's 1m default; and proxies `/status` separately with `access_log off`, because the health check is served ahead of authentication.

The README separates the two credential sets explicitly — `ME_CONFIG_MONGODB_URL` carries the MongoDB credentials, `ME_CONFIG_BASICAUTH_*` gate the web interface only — since conflating them is a recurring support question.

No changes to the Dockerfile, and no lockfiles touched.

**Verification, and its limit:** `docker compose config` resolves and the build context points at the repository root; `nginx -t` passes on the proxy config. I could **not** bring the full stack up on this machine — mongod fails to allocate its journal with the disk at 85% — so this is not verified running end to end. Worth someone doing before merge.

Idea from #1796 by @Mamoona-Ghania, which I closed with the reasoning above.

### DIFF
--- a/examples/docker-compose/README.md
+++ b/examples/docker-compose/README.md
@@ -1,0 +1,49 @@
+# Docker Compose example
+
+Runs MongoDB, mongo-express and nginx together, with nginx as the only published port.
+
+```console
+$ docker compose -f examples/docker-compose/compose.yaml up --build
+```
+
+Then open <http://localhost:8080> and log in with `admin` / `change-me-too`.
+
+Run it from the repository root. The build context in [compose.yaml](./compose.yaml) is
+relative to that file and points back at the repository, so the image is built from the
+working tree rather than pulled — which is what makes this useful for trying out a change.
+
+## What the pieces do
+
+`mongo` starts with a root user and a named volume, so data survives `docker compose down`.
+Its healthcheck gates startup: mongo-express waits for `service_healthy` rather than racing
+the database.
+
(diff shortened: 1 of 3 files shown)


BlackthornYugen (approved):
LGTM, but is it beneficial to deploy it with nginx when using compose?

PR #1819, opened 2026-01-19
feat: implement pluggable auth system and fix OIDC crash (#1733, #1766)

shayandutta:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1819)
- - -

# Implement Form-Based Authentication & Fix OIDC Crash

## Issues Resolved

Closes #1733

Fixes #1766

## Summary

This PR implements a pluggable authentication system using Passport.js that solves two critical issues:

1. **Form-based login for password manager compatibility (#1733)**

2. **Graceful handling of missing OIDC dependencies (#1766)**

## Background: Why Both Issues Are Related

While implementing #1733 (form-based authentication), it became clear that the existing authentication architecture needed refactoring to support multiple strategies cleanly. During this refactoring, I discovered and fixed #1766 (OIDC crash issue) as they both stem from the same root cause: **hardcoded authentication logic** without proper error handling.

The pluggable architecture solves both issues while maintaining 100% backward compatibility.

### New Files

- **`lib/auth.js`** (278 lines): Pluggable authentication module using Passport.js with support for multiple strategies

- **`lib/views/login-form.html`** (116 lines): Modern login page with password manager compatibility

### Modified Files

- **`package.json`**: Added `passport-local` dependency

- **`lib/router.js`**: Integrated auth module, added login/logout routes, proper middleware ordering

- **`config.js` & `config.default.js`**: Added `authStrategy` option with comprehensive documentation

- **`app.js`**: Updated authentication warnings

### Issue #1733: Form-Based Authentication

**Problem:**

Password managers cannot save/auto-fill HTTP Basic Auth credentials. Users had to manually copy/paste credentials every time, making the experience "extremely uncomfortable and tedious."

**Solution:**

Implemented form-based login (`authStrategy: 'form'`) with:

- ✅ Proper HTML form that password managers recognize

- ✅ Autocomplete attributes for seamless auto-fill

- ✅ Modern UI with Bootstrap 5

- ✅ Show/hide password toggle

- ✅ Session-based authentication

- ✅ Logout functionality

- ✅ CSRF protection

**Usage:**

```javascript

// config.js

export default {

  authStrategy: 'form',  // Enable form-based login

  basicAuth: {

    username: 'admin',

    password: 'pass'

  }

}

```

### Issue #1766: OIDC Crash Prevention

**Problem:**

App crashed when OIDC was enabled but `passport-openidconnect` package wasn't installed (common in Docker minimal builds).

**Solution:**

[...]

PR #1902, opened 2026-08-01
chore: pin yarn as the package manager and ignore npm lockfiles

shakaran:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1902)
- - -

The project is already committed to yarn — the release is vendored in `.yarn/releases` and `.yarnrc.yml` points at it — but nothing declares that where tooling can see it, and `package-lock.json` was never ignored.

**That has cost real review time.** Three pull requests carried an accidental npm lockfile:

| PR | `package-lock.json` |
|---|---|
| #1888 | 14,409 lines |
| #1796 | 13,692 lines |
| #1819 | 13,693 lines |

Nearly 42,000 lines of noise, each of which had to be spotted and stripped during review, and each of which buried the actual change — #1819's real diff was 342 lines under 13,693.

#1575 approached the same friction from the other side, proposing to keep `package-lock.json` in step with `yarn.lock` through a `postinstall` hook that runs npm. That cannot work: the hook re-triggers the tool that runs it, which is the loop its author left as an open TODO in June 2024, and `postinstall` in a published package also fires on the machines of anyone installing mongo-express as a dependency — which people do, since it is usable as middleware.

**Two lines close the class instead:**

- `packageManager: "yarn@4.12.0"`, so corepack selects the right tool without anyone having to know
- `package-lock.json` and `npm-shrinkwrap.json` in `.gitignore`

**Verified**, rather than assumed: creating an npm lockfile in the tree no longer shows up in `git status` (`git check-ignore` confirms the rule matches), and `yarn install --immutable` still passes with the version pinned. Lint clean, 135 tests passing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,11 @@ Thumbs.db
 
 # Node and related ecosystem
 # ==========================
+# This project is built with yarn; the vendored release in .yarn/releases and the
+# packageManager field are the source of truth. An npm lockfile here is always accidental,
+# and several pull requests have carried one by mistake.
+package-lock.json
+npm-shrinkwrap.json
 .nodemonignore
 .sass-cache/
 node_modules/
--- a/package.json
+++ b/package.json
@@ -112,6 +112,7 @@
   "engines": {
     "node": ">=22.0.0"
   },
+  "packageManager": "yarn@4.12.0",
   "license": "MIT",
   "scripts": {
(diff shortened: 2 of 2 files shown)


shakaran:
Merged as part of 8527296 (#1908). The content is on master; closing this as superseded rather than leaving a duplicate open.

PR #1903, opened 2026-08-01
fix: build redirect targets from configuration, not from the request

shakaran:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1903)
- - -

Closes the open-redirect alerts on master: **#42** (`lib/routes/database.js:62`, the one raised here) plus **#7**, **#8**, **#9** and **#10** in `lib/router.js`.

All of them are the same shape. `res.locals.baseHref` comes from `buildBaseHref(req.originalUrl, req.url)` — request-derived data reaching `res.redirect`, which is what CodeQL reports as `js/server-side-unvalidated-url-redirection`. It gives the caller a hand in where other people get sent.

All ten redirects across `router.js`, `routes/database.js` and `routes/collection.js` now use `config.site.baseUrl`. The two one-off fixes of the same kind made along the way in #1900 and #1811 are folded into the same helper rather than left as separate spellings.

**Left alone deliberately:** the other uses of `res.locals.baseHref` build hrefs rendered into templates, not `Location` headers. Those should keep reflecting how the app was actually reached, or links break behind a proxy.

**One wrinkle worth knowing:** `config.site.baseUrl` needs the `|| '/'` fallback. Only `config.default.js` supplies a value, and `app.js` patches it in at startup — so `middleware()` called directly, as the tests do, sees `undefined`. Without the fallback this would have shipped `res.redirect(undefined)`.

**Verified the behaviour is unchanged**, comparing targets before and after the change:

| | before | after |
|---|---|---|
| missing database | `/` | `/` |
| missing collection | `/db/<db>` | `/db/<db>` |
| missing document | `/db/<db>/<collection>` | `/db/<db>/<collection>` |

Adds `test/lib/routers/redirectSpec.js`: that a configured `baseUrl` is honoured — which fails if the request-derived value is put back — and that paths shaped like `//evil.example.com` never yield an off-host `Location`.

Lint clean, 141 tests passing.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -63,6 +63,11 @@ const buildId = function (id, query) {
 };
 
 const router = async function (config) {
+  // Redirect targets come from configuration, not from res.locals.baseHref: that is derived
+  // from req.originalUrl, so feeding it to res.redirect hands the caller influence over
+  // where other people get sent. CodeQL reports it as js/server-side-unvalidated-url-redirection.
+  const baseHref = config.site.baseUrl || '/';
+
   // appRouter configuration
   const appRouter = express.Router();
   let mongo = null;
@@ -259,7 +264,7 @@ const router = async function (config) {
     // Make sure database exists
     if (!mongo.connections[id]) {
       req.session.error = 'Database not found!';
-      return res.redirect(res.locals.baseHref);
+      return res.redirect(baseHref);
     }
 
     req.dbName = id;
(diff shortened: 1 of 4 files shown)


shakaran:
Merged as part of dbd4fb2 (#1909). The content is on master; closing this as superseded rather than leaving a duplicate open.

PR #1904, opened 2026-08-01
fix: make the multi-connection config reachable, and document it

### LINKED ISSUE #1787: Add documentation for connecting to multiple databases

shakaran:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1904)
- - -

Closes #1787.

The issue asks for documentation on connecting to more than one database. Writing it turned up why the earlier attempt (#1790) could not be followed by the person who opened the issue: **the feature was unreachable**.

`lib/db.js` has always accepted a list of connections —

```js
const connections = Array.isArray(config.mongodb) ? config.mongodb : [config.mongodb];
```

— and keys databases as `${connectionName}_${dbName}` once there is more than one. But `config.default.js` declares `mongodb` as an object, and two things stood between that and a working setup:

| | |
|---|---|
| `deepmerge` threw `(target || []) is not iterable` | a `config.js` overriding the object with an array made the app refuse to load it: **"Unable to load config.js!"** |
| the startup guard read `config.mongodb.connectionString` | `undefined` on the list form, so it reported no connection string and exited — that one is mine, from #1888 |

Both fixed, and only then documented. The order matters: the previous documentation described something that could not work, which is why the reporter came back saying it did not match `config.default.js`.

**Verified end to end**, running `app.js` against two ephemeral servers with the `config.js` exactly as written in the new README section:

```
ARRANQUE>>> OK
BASES>>> orders_orders reporting_analytics
```

The README section explains the two cases that were being conflated in the issue thread — several databases on one server (`ME_CONFIG_MONGODB_ENABLE_ADMIN`) versus several servers (the list form, which needs a `config.js` because the environment variables describe one connection only) — and the `<connectionName>_<databaseName>` naming, which is otherwise surprising.

Adds `test/lib/multiConnectionSpec.js`: the merge itself, and that databases from both connections are served under their prefixed names. Both fail without the `deepmerge` fix.

`deepmerge` had no test coverage at all before this.

Lint clean, 137 tests passing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,6 +287,41 @@ ME_CONFIG_SITE_BASEURL=/<base-url>
 
 To register your client, you will need the application's redirect URI, which can be obtained by appending `/callback` to the application base URL: Eg. https://example.com/mongo-express/callback
 
+## Connecting to more than one database
+
+A single connection already exposes every database on that server, provided admin access is
+on — set `ME_CONFIG_MONGODB_ENABLE_ADMIN=true`, or `mongodb.admin` in `config.js`. Without
+it, only the database named in the connection string is listed.
+
+To reach databases on **separate MongoDB servers**, give `mongodb` a list of connections
+instead of a single one. This needs a `config.js`: the environment variables describe one
+connection only.
+
+```js
+// config.js
+export default {
+  mongodb: [
+    {
+      connectionString: 'mongodb://user:pass@mongo-a:27017/?authSource=admin',
+      connectionName: 'reporting',
+      admin: true,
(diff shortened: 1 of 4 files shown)


shakaran:
Merged as part of bc03ff6 (#1911). The content is on master; closing this as superseded rather than leaving a duplicate open.

Issue #1712, opened 2025-01-14
Back button on document page doesn't work when ME_CONFIG_OPTIONS_READONLY is true


muu0317:
branch: master
latest commit ID: d832372

If `ME_CONFIG_OPTIONS_READONLY` is true, clicking the "Back" button from document page will return the following error:

<img width="415" alt="Screenshot 2025-01-14 at 17 28 53" src="https://github.com/user-attachments/assets/e542423f-3731-474f-a136-bb9abd1b89f7" />

I tried to set `type="button"` to the button part on document.html and it worked.

lib/views/document.html
Line: 86, 111
ASIS: `<button onclick="return history.back()" class="btn btn-warning btn-large btn-block">`
TOBE: `<button type="button" onclick="return history.back()" class="btn btn-warning btn-large btn-block">`

Could you please check ?

PR #1905, opened 2026-08-01
fix: report an unreachable database, and stop the read-only Back button submitting

### LINKED ISSUE #1671: New version, silent error on connection error
With the new changes to allow to have "login page" now the behavior on start if unable to connect has completely changed.

https://github.com/mongo-express/mongo-express/blob/3f91f4267a7ea50ebbe82216b25c4cadd33cfdf3/lib/router.js#L67

On old versions it will fail, and propagate.
And on the docker containers, crash and that would mark the container as unhealthy.

I would say the old behavior should be maintained as default, because is the known behavior for a lot of people.

But if that is not the case, at least a boolean to have the old behavior.
Let's suppose config.throwErrorOnInitialConnectionFailure
inside the catch, if that is true, just propagate the error instead of a silent failure with log

shakaran:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1905)
- - -

Two issues found while triaging. Closes #1671 and #1712.

## #1671 — the health check said ok while the database was unreachable

```js
try { mongo = await db(config); } catch (error) { console.debug(error); }
```

A startup failure was swallowed into `console.debug`, and the health check returned `ok` unconditionally. Reproduced against an unreachable server:

| | before | after |
|---|---|---|
| `/status` | **200** `{"status":"ok"}` | **503** `{"status":"error","error":"No MongoDB connection"}` |
| log | debug line most people never see | `Could not connect to MongoDB on startup` at error level |

That combination is what the issue is really about: a container with no database reports healthy, so nothing restarts it and nothing takes it out of rotation.

The app still starts, so the runtime connection form can be used to supply working credentials — the check reflects the state rather than forcing an exit.

**Known limit:** this reports whether a connection was ever established. It does not ping on every probe, so a database that disappears later is not detected. Making the check active is a separate decision, since it puts a round trip on every probe.

## #1712 — the read-only Back button also submitted the form

```html
<button onclick="return history.back()" ...>
```

It sits inside the POST form in `document.html`, and a `<button>` without an explicit `type` defaults to `submit`. `history.back()` returns `undefined`, which does not cancel the submission — so the click navigated back *and* posted. Fixed with `type="button"` on both occurrences.

Worth noting the editable path was never affected, which is why the issue only shows up with `ME_CONFIG_OPTIONS_READONLY`: `onBackClick()` returns `false` explicitly.

## Verification

Both regressions are covered, and both new cases fail when the fixes are undone — reverting the two changes takes the suite from 137 passing to 135 passing, 2 failing.

Lint clean.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -4,6 +4,7 @@ import bodyParser from 'body-parser';
 import cookieParser from 'cookie-parser';
 import errorHandler from 'errorhandler';
 import express from 'express';
+import pico from 'picocolors';
 import favicon from 'serve-favicon';
 import logger from 'morgan';
 import methodOverride from 'method-override';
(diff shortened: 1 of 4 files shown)


shakaran:
Merged as part of b86cc3c (#1912). The content is on master; closing this as superseded rather than leaving a duplicate open.

PR #1906, opened 2026-08-01
fix: export the aggregate result, not a broken find()

shakaran:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1906)
- - -

Closes #1674.

All three export endpoints — `export`, `expArr` and `expCsv` — handed `_getQuery()`'s result straight to `collection.find()`. With the **Aggregate** checkbox ticked that result is a pipeline array, and `find()` rejects it:

```
MongoInvalidArgumentError: Query filter must be a plain object or ObjectId
```

Reproduced against a real collection. Worth noting *how* it fails: the error is thrown inside the cursor stream rather than at the call site, so the `try/catch` wrapped around each handler never saw it — the request was left hanging instead of redirecting with a message.

`_getItemsAndCount` has always branched on `runAggregate` for the collection view. The exports never did. They now share `_getExportCursor()`, which runs the pipeline through `aggregate()` — carrying sort and projection as `$sort` and `$project` stages, matching what the view does — or falls back to `find()` exactly as before when the box is unticked.

Three regression tests, one per endpoint. All three fail without the fix (138 passing → 135 passing, 3 failing).

Lint clean.

Issue #1765, opened 2025-06-16
Upload 1.1.0 to Dockerhub


PeterSR:
Hi! 

I am really interested in using the OIDC feature (#1666), but it does not seem to be on Dockerhub.

Would it be possible to publish a new version?

shakaran:
Nobody has spelled out where this is actually stuck, so here is the chain, traced end to end. It is longer than "someone needs to push a button".

**1. There is no 1.1.0 to publish.** This repository is on `1.1.0-rc-4`; the tags stop there and npm's `latest` is `1.1.0-rc-4`. Cutting 1.1.0 — deciding master is ready, bumping the version, tagging — is a maintainer call and has to happen first.

**2. Docker Hub is not published from this repository.** `.github/workflows/publish.yml` publishes to npm and creates a GitHub release. Nothing else. `mongo-express` is a **Docker Official Image**, built from a different repository:

```
# library/mongo-express in docker-library/official-images
GitRepo: https://github.com/mongo-express/mongo-express-docker.git
Maintainers: Nick Cox (@knickers), John Steel (@BlackthornYugen)
```

**3. That repository has no 1.1 at all.** It contains only `1.0/`, with four variants. Someone has to add `1.1` to `versions.json` and generate the Dockerfiles.

**4. And it cannot be added yet.** Two hard dependencies on step 1:

- `versions.sh` picks the upstream version with `^1\.1\.[0-9]*$`, which matches `1.1.0` but **not** `1.1.0-rc-4`.
- `Dockerfile.template` builds from a branch, not a tag: `ARG MONGO_EXPRESS_VERSION=release/v{{ .version }}`. The branches here are `release/v1.0.1`, `release/v1.0.2`, `release/v1.0.3` — there is **no `release/v1.1.0`**, so any 1.1 Dockerfile would fail at `git clone --branch`.

**5. Then a PR to `docker-library/official-images`** to update the manifest.

**Step 3 is already jammed independently.** mongo-express/mongo-express-docker#132 has been open since October 2025 — it drops Alpine 3.18/3.19 and Node 18, all EOL. I reviewed it today: it has aged in the queue and now adds Node 20, which went EOL on 2026-04-30, plus Alpine 3.21, which expires this November. It needs a refresh to Node 22/24 and Alpine 3.22+ before it lands.

---

[...]

PR #1907, opened 2026-08-01
[Release] v1.1.0-rc-5

shakaran:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1907)
- - -

Bumps the version so a release candidate can be cut and published.

**Why an rc and not 1.1.0.** The 1.1 milestone has eight open items, five of which are fixes waiting on review — #1902 through #1906, including the open-redirect work in #1903 that closes five CodeQL alerts. Cutting 1.1.0 now would ship the milestone incomplete, and an npm version cannot be republished once it is out.

**What this unblocks.** #1765 has been open since June asking for a newer image on Docker Hub; the published one is `1.0.2-20-alpine3.19` from November 2024. Of the 150 open issues here, 18 are labelled as old-version problems and 8 report errors from `docker-entrypoint.sh` — a file that only exists in mongo-express-docker. An rc gives something current to build and test the image against.

**Note on the flow, for whoever runs the workflow:** `scripts/release.cjs` passes `prerelease: false`, so this will appear as a normal release on GitHub rather than a pre-release, and `npm publish` without a `--tag` will move `latest` to the rc. That matches how rc-1 through rc-4 were handled, so this changes nothing — but it is worth knowing rather than discovering.

**Still blocked after this**, for the record: `versions.sh` in mongo-express-docker selects upstream versions with `^1\.1\.[0-9]*$`, which does not match an rc. The official image still needs a final 1.1.0; this only unblocks manual builds against the `release/v1.1.0-rc-5` branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.1.0-rc-4",
+  "version": "1.1.0-rc-5",
   "author": "https://github.com/mongo-express",
   "name": "mongo-express",
   "type": "module",


shakaran:
@rtritto @BlackthornYugen — could one of you take a look at this when you get a chance?

I cannot approve or merge my own pull requests, and this one is the last thing standing between the 1.1 milestone and a published release candidate, so it is blocked on a review rather than on any remaining work.

It is a small one: the version bump to `1.1.0-rc-5` and nothing else. The substance already landed in the pull requests merged under the milestone. Once this is in I will run `publish.yml` and cut the `release/v1.1.0-rc-5` branch.

Worth flagging for the release notes: this cycle moves the MongoDB driver to 7.x, which raises the minimum supported server version. Anyone pointing mongo-express at an older MongoDB will find out at runtime, since CI only exercises the versions `mongodb-memory-server` provides.

PR #1914, opened 2026-08-04
fix: import the files this app exports, and the ones other tools write

### LINKED ISSUE #1225: Unable to import its own export file

shakaran:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1914)
- - -

Closes #1225, open since July 2023.

Three defects met in the same place. Between them, the only file that ever imported was a single-line array — this app’s own `Export --jsonArray` output, and nothing else. That matches every report in the thread.

Measured before the change:

| File | Before | After |
|---|---|---|
| Array on one line (`Export --jsonArray`) | 200 | 200 |
| One document per line (mongoexport) | **400** | 200 |
| A single document | **400** | 200 |
| Array pretty-printed (Compass) | **400** | 200 |
| Documents back to back (`Export Standard`) | **400** | 200 |

### Export wrote a format nothing can read

`Export Standard` emitted `{..}{..}{..}` with no separator, on one line. That is not JSON, and not the newline-delimited format `Import --mongoexport json` names, so mongoimport, Compass and this app all rejected it. Each document now ends with a newline.

### Import only ever read one shape

It parsed line by line, so anything pretty-printed failed — hence the workaround in the thread, stripping every line break by hand. It then iterated each parsed line as an array, so a plain document threw `is not iterable` and surfaced as a bare 400.

Parsing now tries the whole file first, covering an array, a single document and a pretty-printed one, then falls back to reading a run of adjacent values, covering mongoexport’s output and the glued-together files earlier versions wrote — so **exports already sitting on disk still import**, which is the literal complaint in the title. The split tracks string state, so a value containing a brace or an escaped quote does not tear in half.

Malformed input is still refused rather than partly imported: content between documents, unbalanced brackets and trailing junk each return 400 and insert nothing.

### Tests

New `test/lib/routers/importSpec.js`: every shape above, the four malformed cases, a round-trip through both export routes, and an assertion that each exported line stands alone as JSON. Suite goes 149 → 162.

Each fix was checked by reverting it alone — the old parser fails 6 of the new tests, and dropping the newline fails the export-format test.

### Separate finding, not fixed here

`ALLOWED_MIME_TYPES` accepts `text/csv`, but there is no CSV parsing, so a CSV passes the type check and then fails as `Bad file content`. Either the importer should read CSV or the type should be dropped; worth its own issue.

### DIFF
--- a/lib/routes/collection.js
+++ b/lib/routes/collection.js
@@ -12,6 +12,119 @@ const ALLOWED_MIME_TYPES = new Set([
   'application/json',
 ]);
 
+/**
+ * Split a run of JSON values that sit next to each other, whether separated by newlines, as
+ * mongoexport writes them, or by nothing at all, which is what this app's own export wrote
+ * before #1225. Brackets are counted outside of strings, so punctuation inside a value never
+ * splits it. Anything between two values other than whitespace means the file is not a run of
+ * documents, and it is rejected rather than silently skipped.
+ */
+const splitAdjacentJsonValues = function (text) {
+  const values = [];
+  let depth = 0;
+  let start = -1;
+  let lastEnd = 0;
+  let inString = false;
+  let escaped = false;
+
+  for (const [index, char] of [...text].entries()) {
+    if (inString) {
+      if (escaped) {
+        escaped = false;
+      } else if (char === '\\') {
+        escaped = true;
+      } else if (char === '"') {
+        inString = false;
+      }
+      continue;
+    }
+
+    if (char === '"' && depth > 0) {
+      inString = true;
+      continue;
+    }
+
+    if (char === '{' || char === '[') {
+      if (depth === 0) {
+        if (text.slice(lastEnd, index).trim() !== '') {
+          throw new SyntaxError('Unexpected content between documents');
+        }
+        start = index;
+      }
+      depth += 1;
+      continue;
+    }
+
+    if (char === '}' || char === ']') {
+      depth -= 1;
+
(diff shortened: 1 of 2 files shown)


shakaran:
@rtritto @BlackthornYugen — a review would be much appreciated whenever you have a moment.

GitHub will not let me approve or merge my own pull requests, so this one is blocked purely on someone else looking at it, not on anything technical: CI is green on Node 22, 24 and 26, CodeQL passes, and the suite goes 149 → 162.

The change is worth a careful eye because it touches how uploaded data is parsed. The parts most worth your attention:

- **The export format changes.** `Export Standard` now separates documents with newlines. It previously wrote them glued together, which no tool could read back — that is the bug — but it is still a change in output, so please sanity-check it against how you use the export.
- **The fallback parser** that reads a run of adjacent JSON values. It tracks string state so a value containing a brace or an escaped quote does not tear in half. There are tests for that, but a second pair of eyes on the scanner is the thing I would most like.
- **Rejection stays strict.** Content between documents, unbalanced brackets and trailing junk all return 400 and insert nothing, so a malformed file cannot import partially.

No rush, and if you would rather I split the export change from the import change into two pull requests, say so and I will.

PR #1915, opened 2026-08-04
fix: stop rejecting uploads over a type the client chose

shakaran:
This change is [Reviewable](https://reviewable.io/reviews/mongo-express/mongo-express/1915)
- - -

Follow-up to #1914, from a finding noted there. **Based on `fix/1225-import-own-export`, so please merge #1914 first** — GitHub will retarget this to `master` automatically.

### The type on an upload was never a control

The importer allowed only `application/json` and `text/csv`. That type is set by the browser and can be anything a client cares to send, so it never decided whether a file was safe — only whether a real one got turned away. Measured against the endpoint:

| Uploaded as | Before | After |
|---|---|---|
| `application/json` | 200 | 200 |
| `text/json` | **400 Bad file** | 200 |
| `text/plain` | **400 Bad file** | 200 |
| `application/octet-stream` | **400 Bad file** | 200 |
| `application/x-json` | **400 Bad file** | 200 |

Several browsers label a `.json` as `text/plain` or `application/octet-stream`, so those users got a bare `Bad file` with nothing to act on. This is the report from @vorou in #1225 that never got explained. What actually validates an import is whether the content parses, and that runs either way.

### CSV was advertised and then failed confusingly

`text/csv` passed the gate and died as unparseable JSON, because no CSV parsing exists. It cannot be added without losing data — the CSV export renders ObjectIds as the text `ObjectId("...")`, flattens nested documents into dotted columns via `flat`, and leaves every value a string, so re-importing would quietly build a collection that no longer matches what was exported.

So a CSV is now refused by name, recognised by type or `.csv` extension, with a message saying to export as JSON instead. Pretending to support it would be worse than saying no.

### Error messages

Parse failures now name the shapes that work instead of only reporting that something was wrong — the thing that had people in #1225 guessing at the format for two years.

### Tests

Suite 162 → 169. Reverting the allow-list fails 5 of the new tests.

### Note for reviewers

This widens what the endpoint accepts, so it deserves a security read. My reasoning is that the multipart content type is attacker-controlled and therefore never provided a boundary; the real ones are unchanged — the upload size limit in `middleware.js`, CSRF, and the read-only guard. If you disagree, I would rather hear it than have this merged on my say-so.

### DIFF
--- a/lib/routes/collection.js
+++ b/lib/routes/collection.js
@@ -7,9 +7,17 @@ import csv from '../csv.js';
 
 const { EJSON } = BSON;
 
-const ALLOWED_MIME_TYPES = new Set([
+/**
+ * A type attached to an upload is whatever the browser decided to send, so it is a hint rather
+ * than a control: any client can set it to anything, and what actually decides the outcome is
+ * whether the content parses. Gating on an allow-list only ever turned away real files — a
+ * .json offered as text/plain or application/octet-stream, which several browsers do send.
+ * These are named only so a CSV can be answered with something better than a parse failure.
+ */
+const CSV_MIME_TYPES = new Set([
   'text/csv',
-  'application/json',
+  'application/csv',
+  'text/comma-separated-values',
 ]);
 
 /**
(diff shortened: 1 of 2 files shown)


shakaran:
@rtritto @BlackthornYugen — review welcome whenever suits you; I cannot approve or merge my own pull requests, so this is blocked on someone else rather than on any remaining work.

This one widens what the import endpoint accepts, so I would value a security read specifically. The reasoning is in the description: the multipart content type is set by the client and so never provided a boundary, while the real limits — upload size, CSRF, the read-only guard — are untouched. If you see it differently, please say so rather than approving on my reasoning alone.

It sits on top of #1914, so that one wants merging first.

emgapps (commented):
PR REVIEWER: NOT READY\n\n[P2] lib/routes/collection.js:778 — CSV metadata is still being used as a rejection control before the content is parsed. A valid JSON document uploaded with text/csv, or under a .csv filename, receives 400 even though the PR’s premise is that client-controlled MIME metadata should not determine acceptance.\n\nPlease attempt JSON/EJSON parsing first. If parsing fails and the MIME type or extension suggests CSV, use that hint to return the more helpful CSV-specific error. Add regression coverage for valid JSON carrying a CSV MIME type and, ideally, a .csv filename.\n\nThe broader MIME allowlist removal looks security-neutral: upload-size limits, CSRF, and read-only enforcement remain intact, and malformed files are fully parsed before insertion. standard-ci passes.\n\nAlso, this PR is based on still-open prerequisite #1914. Merge that PR and retarget/rebase this one before final approval.

